PR #15, opened 2014-09-29
type alias for EdgeInsets on OSX

JaSpa:
if the type alias is used, one can simply pass a `NSEdgeInsets` struct as well. The duplication of `EdgeInsetsMake` and `EdgeInsetsZero` is not needed.


robertjpayne:
I've decided to leave this in there for now, mostly just so OS X matches iOS's API.

PR #27, opened 2014-11-09
Update README.md with import instructions

ndbroadbent:
Not sure if `AppDelegate.swift` is the best place to do this, but it worked for me.


robertjpayne:
@ndbroadbent Snappy is now ready to go with Cocoapods Pre-release. I'd recommend installing that way if you can but I'll continue to find a better way to integrate for iOS targets that do not support frameworks.

PR #35, opened 2015-01-03
Make the Snappy scheme shared.

danieleggert:
This allows using [Carthage](https://github.com/Carthage/Carthage) to add Snappy to a Cocoa application.


devxoul:
+1. This will be needed for CI.

robertjpayne:
Thanks!

PR #36, opened 2015-01-05
Add `updateConstraints`.

devxoul:
Concept
-------

#### 1. Just update a constant

If the new constraint is exactly the same with existing, `updateConstraints` will just update a constant of existing constraint.

```swift
view.snp_makeConstraints { make in
    make.top.left.right.and.height.equalTo(10)
    return // explicit return due to a swift bug
}

view.snp_updateConstraints { make in
    make.height.equalTo(10) // just update constant
    return
}
```

#### 2. Reinstall

If the new constraint has different relation from existing, `updateConstraints` will uninstall existing and install the new one.

```swift
view.snp_makeConstraints { make in
    make.top.left.right.and.height.equalTo(10)
    return
}

view.snp_updateConstraints { make in
    // `secondItem` has changed, so this constraint will be re-installed.
    make.top.equalTo(otherView.snp_bottom).with.offset(100)
    return
}
```

Animation
---------

Animation became easy.

```swift
view.snp_makeConstraints { make in
    make.left.equalTo(0)
    return
}

view.snp_updateConstraints { make in
    make.left.equalTo(100)
    return
}
view.setNeedsLayout()
UIView.animateWithDuration(0.5, animations: view.layoutIfNeeded)
```

### DIFF
--- a/Snappy/Constraint.swift
+++ b/Snappy/Constraint.swift
@@ -45,6 +45,8 @@ public class Constraint {
     
     public var and: Constraint { return self }
     public var with: Constraint { return self }
+
+    public var updateExisting: Bool = false
     
     // MARK: initializer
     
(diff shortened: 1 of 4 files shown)


robertjpayne:
Thanks for this PR I used it as inspiration but I implemented it a bit different; more similar to the way Masonry handles it by re-using the actual NSLayoutConstraint object.

Anyways, this is now available on the develop branch!

devxoul:
Thanks. Do you have an specific date plan for support `updateConstraints`?

robertjpayne:
@devxoul it's now available in master, it may have a few quirks but I tried to implement it as close to Masonry's behaviour as possible.

devxoul:
Great. I just saw it on CocoaPods. :beers:

devxoul:
@robertjpayne What do you think of **Concept 2**? Masonry-like implementation supports only **Concept 1** but this PR contains both. I think many projects need **Concept 2** approach.

PR #39, opened 2015-01-12
Replace constraints when update to different relationship.

devxoul:
There are needs to replace existing constraints to another relationship.

For example:

<sup>[Image from Stack Overflow - "How to use auto-layout to move other views when a view is hidden?"](http://stackoverflow.com/questions/18065938/how-to-use-auto-layout-to-move-other-views-when-a-view-is-hidden)</sup>

We can make `UIView`'s width to 0, but it's not a good solution if there are lots of UI elements in the screen and have various cases.

In this case, the most expected code should be:

```swift
label1.snp_makeConstraints { make in
    make.left.equalTo(view.snp_right).with.offset(10)
}

label1.snp_updateConstraints { make in
    make.left.equalTo(10)
}
```

`updateConstraints()` currently provides updating **constant** only, so this code wouldn't be executed.

This PR makes `updateConstraints()` to replace existing constraints if needed. Some test codes are included.

### DIFF
--- a/Snap.xcodeproj/project.pbxproj
+++ b/Snap.xcodeproj/project.pbxproj
@@ -7,6 +7,16 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0364732D1A64086800E1F188 /* UpdateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0364732C1A64086800E1F188 /* UpdateTests.swift */; };
+		0364732E1A6408F200E1F188 /* SnapTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EE91728C19CB304E007888CF /* SnapTests.swift */; };
+		0364732F1A64090E00E1F188 /* View+Snap.swift in Sources */ = {isa = PBXBuildFile; fileRef = EEBCC9F119CC65040083B827 /* View+Snap.swift */; };
+		036473301A64090E00E1F188 /* ConstraintMaker.swift in Sources */ = {isa = PBXBuildFile; fileRef = EEBCC9FB19CC65430083B827 /* ConstraintMaker.swift */; };
+		036473311A64090E00E1F188 /* Constraint.swift in Sources */ = {isa = PBXBuildFile; fileRef = EEBCC9FF19CC66020083B827 /* Constraint.swift */; };
+		036473321A64090E00E1F188 /* ConstraintItem.swift in Sources */ = {isa = PBXBuildFile; fileRef = EEBCC9F719CC65260083B827 /* ConstraintItem.swift */; };
+		036473331A64090E00E1F188 /* ConstraintAttributes.swift in Sources */ = {isa = PBXBuildFile; fileRef = EEBCC9F319CC65110083B827 /* ConstraintAttributes.swift */; };
+		036473341A64090E00E1F188 /* ConstraintRelation.swift in Sources */ = {isa = PBXBuildFile; fileRef = EEBCC9F519CC65200083B827 /* ConstraintRelation.swift */; };
+		036473351A64090E00E1F188 /* LayoutConstraint.swift in Sources */ = {isa = PBXBuildFile; fileRef = EEBCC9FD19CC65510083B827 /* LayoutConstraint.swift */; };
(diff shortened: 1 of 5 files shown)


robertjpayne:
@devxoul this is a good idea, but at the moment I'm trying to match Masonry's logic for updating which is to only update the constant.

I'm debating an add-on to Snap that allows for tagging constraints on creation and turning them on/off by tags similar to CSS classes.

The reasoning behind this is to keep updateConstraints simple and stupid while allowing a better and less ambiguous API for replacing/enabling/disabling pre-setup constraints.

devxoul:
@robertjpayne Thanks. Tagging like CSS class looks great, but can't we use `NSLayoutConstraint`'s `active` property that introduced at iOS 8?
> The receiver may be activated or deactivated by manipulating this property.  Only active constraints affect the calculated layout.  Attempting to activate a constraint whose items have no common ancestor will cause an exception to be thrown.  Defaults to NO for newly created constraints.

robertjpayne:
@devxoul yup the tagging API will use the active property if possible ( not running on iOS 7 target ).

There will be a few new API's on UIView/NSView:

```swift
// will create a set of constraints but not install them
snp_prepareConstraints(tag: String, block: (maker: ConstraintMaker) -> ())

// will install constraints found with this tag if not already installed
snp_enableConstraints(tag)

// will uninstall constraints found with this tag if installed
snp_disableConstraints(tag)

// will toggle ( install if uninstalled and uninstall if installed ) constraints found with this tag
snp_toggleConstraints(tag)

// will uninstall if installed and remove constraints found with this tag
snp_removeConstraints(tag)
```

Expanding on this API I want to add a layout API to view controllers so one can control larger portions of constraints at a view controller level. It will be similar to the view's API but be a LayoutMaker which will allow API like `view ~> make.width.equalTo(50)` so you can setup more than one view in the block.

cloudkite:
@robertjpayne the way `mas_updateConstraints` works is pretty confusing.
What you are proposing sounds alot better.

I'd like to suggest a slightly different approach. I'm not super keen on using strings to tag/group objects in code.

How about if `makeConstraints` returned a composite constraint which held all the individual constraints contained within an array but conformed to the same api as a single `Constraint`

then you could have something like
```
self.view1ConstraintGroup = view1.snp_prepareConstraints { make in
  make.top.equalTo(superview.snp_top)..offset(padding.top)
  make.left.equalTo(superview.snp_left)..offset(padding.left)
  self.rightConstraint = make.right.equalTo(superview.snp_right)..offset(padding.right)
}

// I want to update all offsets
self.view1ConstraintGroup.offset(20);
// but I want right to be different
self.rightConstraint.offset(10);

// I want to deactivate all constraints
self.view1ConstraintGroup.activate();

// I want to uninstall all constraints
self.view1ConstraintGroup.uninstall();

// I want to reinstall only right constraint
self.rightConstraint.install();
```

What are your thoughts?

cloudkite:
It could also open up the possibility of sub-groupings like

self.view1ConstraintGroup = view1.snp_prepareConstraints { make in
  make.top.equalTo(superview.snp_top).offset(padding.top)
  self.view1Sides = make.group {
    make.left.equalTo(superview.snp_left).offset(padding.left)
    make.right.equalTo(superview.snp_right).offset(padding.right)
  }
}

robertjpayne:
@cloudkite yea that's pretty reasonable, I suppose I hate storing constraints on variables even in swift. It's pretty annoying.

I also need to refine the API so you can do more than 1 view at a time in a grouping.

cloudkite:
I guess with a grouping approach you could still use strings instead of variables by shoving the result into a dictionary.

ie

```swift
var constraintGroups = [String: Constraint]()
constraintGroups["view1"] = view1.snp_prepareConstraints { make in
  make.top.equalTo(superview.snp_top).offset(padding.top)
  make.left.equalTo(superview.snp_left).offset(padding.left)
  make.right.equalTo(superview.snp_right).offset(padding.right)
}
```

or use the actual view as the key

```swift
var constraintGroups = [UIView: Constraint]()
constraintGroups[view1] = view1.snp_prepareConstraints { make in
  make.top.equalTo(superview.snp_top).offset(padding.top)
  make.left.equalTo(superview.snp_left).offset(padding.left)
  make.right.equalTo(superview.snp_right).offset(padding.right)
}
```

robertjpayne:
The latest `develop` branch now contains support for `snp_prepareConstraints`. It is up to the user to store the constraints and add/remove them as they see fit.

PR #46, opened 2015-02-04
Fix for typo in OS check enabling iOS specific layout attributes

sebreh:
Fixes a typo that caused iOS-specific layout attributes not to be included installing a constraint. I discovered this when I noticed aligning to leading margin to a super view fell back to using the layout attribute of the first item of the constraint.


devxoul:
OMG, my fault in #28...

sebreh:
Cool, could you merge this PR? Great library btw.

devxoul:
I'm not a committer, we should ask to @robertjpayne.

robertjpayne:
Thanks @sebreh I'll get Cocoapods released soon(ish).

sebreh:
👍

PR #50, opened 2015-02-10
Workaround to fix Swift 1.2 regression

Reflejo:
Since we can't tell when this is going to be fixed in the language, It'd be great if Snap can workaround this issue in order to be able to use Snap on Swift 1.2.

### DIFF
--- a/Snap/Constraint.swift
+++ b/Snap/Constraint.swift
@@ -411,7 +411,7 @@ public class Constraint {
             }
             
             // clean up the snp_installedLayoutConstraints
-            var layoutConstraints = view.snp_installedLayoutConstraints
+            var layoutConstraints = Array(view.snp_installedLayoutConstraints)
             var layoutConstraintsToKeep = Array<LayoutConstraint>()
             for layoutConstraint in layoutConstraints {
                 if !contains(layoutConstraintsToRemove, layoutConstraint) {
--- a/Snap/Constraint.swift
+++ b/Snap/Constraint.swift
@@ -402,10 +402,10 @@ public class Constraint {
         if let view = self.installedOnView {
             // remove all installed layout constraints
             var layoutConstraintsToRemove = Array<LayoutConstraint>()
-            if let installedLayoutConstraints = self.installedLayoutConstraints?.allObjects as? Array<LayoutConstraint> {
+            if let installedLayoutConstraints = self.installedLayoutConstraints?.allObjects.map({ $0 as! LayoutConstraint }) {
                 layoutConstraintsToRemove += installedLayoutConstraints
             }
-            
+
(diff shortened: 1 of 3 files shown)


ryanmaxwell:
@Reflejo A repeated remake will crash, e.g.

```objc
let box = UIView()
        
        view.addSubview(box)
        
        box.snp_remakeConstraints { make in
            make.edges.equalTo(self.view)
        }
        
        box.snp_remakeConstraints { make in
            make.edges.equalTo(self.view)
        }

```

Reflejo:
@ryanmaxwell Try the PR again please, ce31f2c29319271ac93bb83da4ba34b44b707401 should fix it.

robertjpayne:
@Reflejo @ryanmaxwell this PR contains 1.2 specific logic specifically the `as!` keyword/cast operator does exist in Swift 1.0.

ryanmaxwell:
@robertjpayne Is there a way to version a pod to a swift language version? any compiler flags?

robertjpayne:
@ryanmaxwell not that I'm aware of. `#defines` are pretty impossible to use now that they are compiled into a secondary framework.

robertjpayne:
@ryanmaxwell just looked too no ENV variables are set during compile time either so out of luck there.

robertjpayne:
Anyways, I'll investigate more into this, potentially I could just keep a branch that tracks for Swift 1.2+ until it's a public release. It'd be really awesome to uncover a way that can automatically do this though.

ryanmaxwell:
@Reflejo https://github.com/Masonry/Snap/commit/ce31f2c29319271ac93bb83da4ba34b44b707401 "fixes" it

robertjpayne:
@ryanmaxwell yea it fixes it, but again just want to make sure you know that it doesn't solve the Swift 1.2 syntax issue.

ryanmaxwell:
@robertjpayne yep I agree a 1.2 swift branch would be helpful for beta users, if we can choose that in the podfile, and keep master supporting release versions.

Reflejo:
I agree that it should live in another branch *for a while*. @robertjpayne if you create that branch I can do a new PR. Also I'd do a new PR adding  the new `@noespace` to the `snp_(re)(make|update)Constraints` methods

robertjpayne:
@ryanmaxwell @Reflejo appears that the crash was just related to the computed property and it's return value.

I've added a `swift-1.2` branch which should fix all of this. I added the `@noescape` attributes but for now they cause SourceKit to go mad and crash A LOT so I'm leaving them out.

robertjpayne:
Actually this can now be downloaded if you point your Cocoapod to `'Snap', :git => 'https://github.com/Masonry/Snap.git'` as it turns out the fixes are compatible with current versions of Swift as well.

PR #51, opened 2015-02-10
Faster closest commonsuperview

Reflejo:
The general idea is to move from an O(n^2) algorithm to a O(N) where N is the height of the view hierarchy tree. The algorithm is very simple: traverse the tree up starting from both views and keep the visited views in a set. When any of the nodes was already visited before: that must be the common superview.

### DIFF
--- a/Snap/Constraint.swift
+++ b/Snap/Constraint.swift
@@ -507,7 +507,7 @@ public class Constraint {
                     return view
                 }
                 views.insert(view)
-                fromView = view.superview
+                toView = view.superview
             }
         } while (fromView != nil || toView != nil)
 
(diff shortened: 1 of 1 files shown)


robertjpayne:
@Reflejo 
How necessary is this? Also what's the performance of the contains lookup? In most cases I imagine constraints are only going up a few layers though it could be more I'm hesitant to integrate this as I don't have tests in place to validate it always works and the widespread issues it would cause in cases of failure.

Reflejo:
Sets lookups are supposed to be O(1) avg as they are backed up on a hashtable.

I can't prove correctness on the code itself but we can't neither for the old implementation. This method is being called for each set of constraints you add to the maker object, which makes the old code O(M  * N ^ 2), but yes, usually n and m are small. The only reason I'm purposing the change is because I think that this algo is very simple and doesn't seem to be risky, but I understand your hesitance.

Reflejo:
This is, again, swift 1.2 specific since I'm using Set(s). So if you are ok with this, it should go to the 1.2 branch.

robertjpayne:
Can you open this as a new PR against the swift-1.2 branch please?

PR #55, opened 2015-02-14
Faster closest commonsuperview

Reflejo:
The general idea is to move from an O(n^2) algorithm to a O(N) where N is the height of the view hierarchy tree. The algorithm is very simple: traverse the tree up starting from both views and keep the visited views in a set. When any of the nodes was already visited before: that must be the common superview.

More info: #51

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ project.xcworkspace
 xcuserdata
 Examples/
 .DS_Store
+Gemfile
+Gemfile.lock
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
-Preview
+Beta
 =======
 
-Snap is currently in preview and regular Changelogs will be provided as it comes out of preview into production quality code.
+# 0.0.6 - February 11th, 2015
+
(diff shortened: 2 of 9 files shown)


robertjpayne:
Hey @Reflejo believe this contains a typo in the do while loop looks like two `fromView = view.superview`?

Reckon there is much downside to using NSSet vs Set? We could then put this into develop?

Reflejo:
@robertjpayne Ok, I just changed `Set<>` to `NSMutableSet`. But I don't see any typo.

By the way, Is there any reason you want to support Swift 1.0+ ?. I think that moving forward, it'd be very beneficial to incorporate all the changes of a fast evolving language like Swift.

Reflejo:
Closing this. I'm gonna create a new one against `develop`

robertjpayne:
@Reflejo there was a typo in the commit I looked at but you had fixed it in a subsequent one my mistake.

I want to maintain Swift 1.0 support until Swift 1.2 is out unless we have really good reason not to. Mostly just so we don't have to maintain two branches separately. I'm all for ditching NSSet and sorts as soon as Swift 1.2 is released but it's quite likely not going to be a public release until May or later.

PR #65, opened 2015-04-02
Set iOS deployment target to iOS 8.0, fixes App Store submission bug

sebreh:
I ran into an issue when submitting a version of my app using Snap to the app store.

screen_shot_2015-04-02_at_12_49_49

Apparently this issue is due to the library using iOS 7 as the minimum deployment target, as described here (Alamofire had the same issue): https://github.com/Carthage/Carthage/issues/351

Since dynamic frameworks are only supported on iOS 8 and up, this PR changes the deployment target on the project level.

I have verified that this does fix the submission bug and the .IPA is accepted by the App Store.


robertjpayne:
Thanks for this!

Nonnus:
actually i am still having this issue, 
even though i am downloading latest revision of alamofire,
is there any solution for this problem ?

ps: i am also getting a alamofile related error regarding unsupported architecture (x86_64, i386) 

screen shot 2015-05-03 at 23 42 50

tc:
@Nonnus did you get a solution to your unsupported architecture (x86_64, i386)  problem?

robertjpayne:
@Nonnus @tc I've never really seen this happen -- I had a ton of problems pre Swift 1.2 but since then all pretty smoothed out.

Mostly I'd attribute these things to Xcode or Cocoapods and recommend using either Carthage or manual integration.

RajvinderMonotype:
@Nonnus did you find any solution? I am getting same issues..

sferrini:
@robertjpayne I've just created a new project, added a `Cartfile` added `SnapKit` as a dependency and runned `carthage update`. I'm still having the @Nonnus and @tc issues.

screen shot

Do you have any idea?

robertjpayne:
@sferrini Few things:

1. What version of Xcode are you using?
2. Are you building for iOS or OS X?
3. What version of SnapKit is actually being checked out/installed?
4. Are you uploading Bitcode?
5. Did you ensure to configure the copy frameworks step?

sferrini:
@robertjpayne :

1. Version 7.2.1
2. iOS
3. 0.19.1
3. Nope, I've removed the :white_check_mark: 
5. Of which steps are you talking about? I've followed the `Carthage` section of http://snapkit.io/docs/

robertjpayne:
@sferrini there is some general carthage build steps you have to use so it copys and codesigns the frameworks properly I thought? I don't use Carthage myself.

sferrini:
@robertjpayne yes there are: https://github.com/Carthage/Carthage#if-youre-building-for-ios-tvos-or-watchos and I just did them again, now it seems to work :+1:

robertjpayne:
@sferrini :+1: glad to hear, unsure what the main cause was!

PR #68, opened 2015-04-10
Fixed bug that leaked every constant added.

Reflejo:
Also fixed some O(N^2) queries to run O(N) and cleanup a little bit the code to support Swift 1.2.

I'll add comments on most of my changes so you understand the reasoning behind the changes.

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -313,7 +313,7 @@ public class Constraint {
                 NSException(name: "Cannot Install Constraint", reason: "Already installed on different view.", userInfo: nil).raise()
                 return []
             }
-            return (self.installedLayoutConstraints?.allObjects as? Array<LayoutConstraint>)!
+            return self.installedLayoutConstraints?.allObjects as! Array<LayoutConstraint>
         }
         
         var newLayoutConstraints = Array<LayoutConstraint>()
@@ -362,7 +362,7 @@ public class Constraint {
         // special logic for updating
         if updateExisting {
             // get existing constraints for this view
-            let existingLayoutConstraints = reverse(layoutFrom!.snp_installedLayoutConstraints)
+            let existingLayoutConstraints = Set<LayoutConstraint>(layoutFrom!.snp_installedLayoutConstraints)
             
             // array that will contain only new layout constraints to keep
             var newLayoutConstraintsToKeep = Array<LayoutConstraint>()
@@ -373,16 +373,8 @@ public class Constraint {
                 var updateLayoutConstraint: LayoutConstraint? = nil
                 
                 // loop through existing and check for match
-                for existingLayoutConstraint in existingLayoutConstraints {
-                    if existingLayoutConstraint == layoutConstraint {
-                        updateLayoutConstraint = existingLayoutConstraint
-                        break
-                    }
-                }
-                
-                // if we have existing one lets just update the constant
-                if updateLayoutConstraint != nil {
-                    updateLayoutConstraint!.constant = layoutConstraint.constant
+                if let index = existingLayoutConstraints.indexOf(layoutConstraint) {
+                    existingLayoutConstraints[index].constant = layoutConstraint.constant
                 }
                 // otherwise add this layout constraint to new keep list
                 else {
@@ -407,9 +399,7 @@ public class Constraint {
         }
         
         // store the layout constraints against the installed on view
-        var layoutConstraints = Array<LayoutConstraint>(layoutFrom!.snp_installedLayoutConstraints)
-        layoutConstraints += newLayoutConstraints
-        layoutFrom!.snp_installedLayoutConstraints = layoutConstraints
+        layoutFrom!.snp_installedLayoutConstraints += newLayoutConstraints
         
         // return the new constraints
         return newLayoutConstraints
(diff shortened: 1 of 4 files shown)


Source/Constraint.swift
```diff
@@ -362,7 +362,7 @@ public class Constraint {
         // special logic for updating
         if updateExisting {
             // get existing constraints for this view
-            let existingLayoutConstraints = reverse(layoutFrom!.snp_installedLayoutConstraints)
+            let existingLayoutConstraints = Set<LayoutConstraint>(layoutFrom!.snp_installedLayoutConstraints)
```
Reflejo: This is just cosmetic and it's for simplification purposes. This avoids the next for-loop.

Source/Constraint.swift
```diff
@@ -407,9 +399,7 @@ public class Constraint {
         }
         
         // store the layout constraints against the installed on view
-        var layoutConstraints = Array<LayoutConstraint>(layoutFrom!.snp_installedLayoutConstraints)
-        layoutConstraints += newLayoutConstraints
-        layoutFrom!.snp_installedLayoutConstraints = layoutConstraints
+        layoutFrom!.snp_installedLayoutConstraints += newLayoutConstraints
```
Reflejo: I believe this was a hack to make beta2 work. This is already fixed in Xcode GM.

Source/Constraint.swift
```diff
@@ -417,27 +407,15 @@ public class Constraint {
...
             }
-            
-            // clean up the snp_installedLayoutConstraints
-            var layoutConstraints = view.snp_installedLayoutConstraints
-            var layoutConstraintsToKeep = Array<LayoutConstraint>()
-            for layoutConstraint in layoutConstraints {
-                if !contains(layoutConstraintsToRemove, layoutConstraint) {
-                    layoutConstraintsToKeep.append(layoutConstraint)
+
+            if let installedConstraints = self.fromItem.view?.snp_installedLayoutConstraints {
```
Reflejo: This!. On the old code, the constraints were trying to be removed based on `view.snp_installedLayoutConstraints` which wasn't doing anything. This fixes a constraint leak that exponentially slows down a view with many `remakeconstraint`s

Source/Constraint.swift
```diff
@@ -417,27 +407,15 @@ public class Constraint {
...
-            // remove all installed layout constraints
-            var layoutConstraintsToRemove = Array<LayoutConstraint>()
-            if let allObjects = self.installedLayoutConstraints?.allObjects {
-                if let installedLayoutConstraints = allObjects as? Array<LayoutConstraint> {
-                    layoutConstraintsToRemove += installedLayoutConstraints
-                }
-            }
-            
-            if layoutConstraintsToRemove.count > 0 {
+            if let layoutConstraintsToRemove = self.installedLayoutConstraints?.allObjects {
```
Reflejo: No need for the previous O(N) operation (that actually runs for each constant removal making it (N^2). I know the constant won't be a lot but removing this actually simplifies the code.

Source/Constraint.swift
```diff
@@ -509,22 +487,22 @@ public class Constraint {
     }
     
     private class func closestCommonSuperviewFromView(fromView: View?, toView: View?) -> View? {
-        var views = NSMutableSet()
+        var views = Set<View>()
```
Reflejo: Bridging is slow. Let's use Swift constructors now that the GM is out

Source/Constraint.swift
robertjpayne: Yea that was a hack.

Source/Constraint.swift
robertjpayne: Think I had added this in due to some Beta 2 crashes but I forget really.

Source/Constraint.swift
robertjpayne: Are we sure that Set is actually using proper equality checks in Swift? The equality check required to compare constraints is an operate overload (See the bottom of the file).

I'm concerned that this wont actually updating existing constraints which are "identical" but different pointers.

robertjpayne:
@Reflejo it would be great if you could try to split up your PR's into more commits ( IE modernizing syntax could be on commit while perf optimisations could be another ).

This would help with merging as while I agree with most of the modernisation I'm more willing to delay merging the performance optimisations until I have proper time to test etc…

Source/Constraint.swift
Reflejo: Regarding to your question: Yes Swift uses the proper equality check. Actually `Hashable` implements `Equatable` because of this. 

Having said that when the object that you insert into the set is an `NSObject` and not a swift object, `isEqual` gets called instead of the overload operator. I'm gonna revert this one.

Source/Constraint.swift
```diff
@@ -362,7 +362,7 @@ public class Constraint {
         // special logic for updating
         if updateExisting {
             // get existing constraints for this view
-            let existingLayoutConstraints = reverse(layoutFrom!.snp_installedLayoutConstraints)
```
Reflejo: @robertjpayne Quick question before I change this implementation (and split the PRs): What is the reasoning for `reverse` this array here?

Reflejo:
Closing this one in favor of the others

PR #70, opened 2015-04-11
Fixed constraints leak + performance optimizations

Reflejo:
This is probably the most urgent one.

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -407,9 +407,7 @@ public class Constraint {
         }
         
         // store the layout constraints against the installed on view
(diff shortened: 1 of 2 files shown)


robertjpayne:
These updates should be in develop now, I sort of manually merged because I'm working on some other things and am shifting some stuff around.

Reflejo:
Nice. :100:

robertjpayne:
Also I flagged everything as final as from my understanding it can improve performance.

robertjpayne:
@Reflejo actually just fixed another bug that was still causing the constraints to be around forever <eek>.

The leak you found I think wasn't too major ( mostly only bad for views that live forever ). As soon as all views involved with a constraint bunch ( basically UIViewController and down ) it would definitely clean up.

All fixed now either way on develop!

PR #75, opened 2015-04-11
Initial Implementation of protocol API

robertjpayne:
This is a first run at a protocol based API chaining. The major changes are:

* Each step of the constraint making process is restricted to a subset of API's to help prevent errors
* and/with have been removed since they were semantic only and complicate things
* To store a reference to a constraint for later usage you need to get a reference from the chain for example: `self.topConstraint = make.top.equalTo(50).priorityHigh().constraint`
* You cannot combine a constant equalTo with an offset anymore. For example `make.top.equalTo(50).offset(50)` is invalid. You should instead just make it `equalTo(100)`
* You can no longer repeat any invalid API's in the chain such as `equalTo(50).offset(50).offset(100)` used to work and result in the `offset(100)` overriding the `offset(50)`. This will now throw a compiler error.

@cloudkite very keen on your thoughts, know this is a big deviation from Masonry


cloudkite:
@robertjpayne this looks great!!

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ CHANGELOG
 # 0.10.0 - Next Release In Development
 
 * **BREAKING:** Using `.equalTo(<EdgeInsets>)` will no longer multiply the constants on the right/bottom edge by -1. This does not affect `.offset(<EdgeInsets>)`
```
cloudkite: Curious what the reasoning behind this is?

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ CHANGELOG
...
+* **BREAKING:** The constraint making chain now utilises protocols to tighten the API's to avoid user error. This may break some syntaxes.
+* **BREAKING:** Semantic `and` and `with` chain variables were removed
```
cloudkite: :+1:

CHANGELOG.md
robertjpayne: Hm good point, I'll revert this as it actually doesn't do what I thought it was going to.

robertjpayne: It's actually quite weird we even allow the equalTo to be UIEdgeInsets because I was thinking that doing edges.equalTo(UIEdgeInsetsMake(10,10,20,20)) would lock top = 10, left = 10, bottom = 20, right = 20 but it actually insets it from the superview by that amount.

PR #78, opened 2015-04-13
Replace `(make) -> Void in ...` with `make in ...` on README

devxoul:
We love shorter.


robertjpayne:
@devxoul while I appreciate the amendment I actually purposely did this as Xcode's autocomplete tools will automatically create the long form so I'm trying to reduce confusion in the docs.

devxoul:
Agree.

Issue #87, opened 2015-05-03
Application unable to submit to App Store


jurvis:
Application couldn't be submitted to the App Store for the following errors:

```
ERROR ITMS-90362: "Invalid Info.plist value. The value for key 'MinimumOSVersion' in bundle ... is invalid. The minimum value is 8.0
ERROR ITMS-90209: "Invalid Segment Alignment. The app binary at ... does not have proper segment alignment. Try rebuilding the app with the latest Xcode version.
ERROR ITMS-90125: "The binary is invalid. The encryption info in the LC_ENCRYPTION_INFO load command is either missing or invalid, or the binary is already encrypted. This binary does not seem to have been built with Apple's linker.
```

I think the best course of action forward will be to change the deployment version to 8.0. Since apps with a deployment target of 7.0 won't be able to submit the app to Apple with Frameworks. Seems pretty close to the issue mentioned here https://github.com/Alamofire/Alamofire/issues/352

Have submitted a PR https://github.com/SnapKit/SnapKit/pull/88 do feel free to merge if it makes sense.

PR #88, opened 2015-05-03
switched deployment target to iOS 8.0 for target and tests

jurvis:
Setting the deployment target as 7.0 prevents the app from being able to be submitted to the App Store with the following errors: 

```
ERROR ITMS-90362: "Invalid Info.plist value. The value for key 'MinimumOSVersion' in bundle ... is invalid. The minimum value is 8.0
ERROR ITMS-90209: "Invalid Segment Alignment. The app binary at ... does not have proper segment alignment. Try rebuilding the app with the latest Xcode version.
ERROR ITMS-90125: "The binary is invalid. The encryption info in the LC_ENCRYPTION_INFO load command is either missing or invalid, or the binary is already encrypted. This binary does not seem to have been built with Apple's linker.
```

Issue raised: https://github.com/SnapKit/SnapKit/issues/87
Thanks for the awesome work!


robertjpayne:
@jurvis just out of curiosity, how are you integrating the framework with your app?

jurvis:
@robertjpayne current using Carthage to do that

robertjpayne:
@jurvis awesome and how are you embedding it into your target? Are you just adding them to the embedded frameworks or are you using the carthage copy-frameworks command?

Only reason I'm trying to narrow this down is because this would essentially kill using Carthage for enterprise apps that can use frameworks on iOS 7.0 just fine.

jurvis:
@robertjpayne cool, good thinking. Didn't think about that bit on enterprise apps.

I'm currently using the carthage copy-frameworks command and adding the .framework to Linked Frameworks and Libraries (just as what Carthage instructs in the README)

robertjpayne:
@jurvis ok, well with that in mind I'll agree and bump this as I'd rather support the majority of users rather than the enterprise minority.

PR #108, opened 2015-06-09
[WIP] Support Swift 2.0

rpowelll:
This pull request should make SnapKit work with the Swift 2.0 beta that was released at WWDC 2015. I'd advise against merging it in to master just yet, as it will break compatibility with Swift 1.3.

### DIFF
--- a/SnapKit.xcodeproj/project.pbxproj
+++ b/SnapKit.xcodeproj/project.pbxproj
@@ -279,7 +279,8 @@
 		DDC9FD8D1981B4DD009612C7 /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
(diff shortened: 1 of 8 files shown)


robertjpayne:
@rpowelll thanks! I'll have a look at this in the next week or so -- probably will maintain a Swift 2.x branch until Swift 2.x is public.

Also a lot of new auto layout stuff to delve into!

rpowelll:
Yep, once I've dug into that more myself I'd be happy to help with some pull requests to support iOS 9 / OS X El Capitan

robertjpayne:
@rpowelll I really appreciate this based a new `swift-2.0` branch off most of these changes. Lets keep any future PR's against it!

PR #114, opened 2015-06-17
Swift 2.0

robertjpayne:
Will leave this PR open until Swift 2.0 launches and it can track changes to make SnapKit compatible.


keith:
Now that the GM is released will we be able to merge this PR soon?

robertjpayne:
Yup! Will be doing it soon :+1:

thedamfr:
Does that means that 14.0 support swift 2.0 ?

PR #115, opened 2015-06-21
Simplified UIViewController extension for layout guides

raylillywhite:
As far as I know, there's no case where you should be using the top of the topLayoutGuide or the bottom of the bottomLayoutGuide, is there?

### DIFF
--- a/Source/ViewController+SnapKit.swift
+++ b/Source/ViewController+SnapKit.swift
@@ -30,17 +30,10 @@ public typealias ViewController = UIViewController
 */
 public extension UIViewController {
     
-    /// top layout guide top
-    public var snp_topLayoutGuideTop: ConstraintItem { return ConstraintItem(object: self.topLayoutGuide, attributes: ConstraintAttributes.Top) }
+    /// top layout guide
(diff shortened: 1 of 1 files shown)


robertjpayne:
@raylillywhite there may be in the future, while typically the top of the top and bottom of the bottom would always be the view that may not be the case always.

Are you experiencing any oddities by using either of these or just feel it's ambiguous/duplicated?

raylillywhite:
No, I'm not experiencing any problems (besides that I autocompleted to the wrong one without noticing). I just thought that the fact that these layout guides have a non-zero height was merely an implementation artifact, and when I was first learning auto layout, it was confusing that I had to specify that I was pinning something to the _bottom_ of the top layout guide. So I thought that removing that distinction would be a helpful abstraction, and I couldn't think of any way that the top of the topLayoutGuide would be meaningful.

I understand that this breaks backwards compatibility, so that alone might be enough reason not to merge this. But if you agree that the simplification could be worthwhile, let me know and I'll resolve the test failures. Otherwise, feel free to close this.

robertjpayne:
@raylillywhite I'm going to go ahead and leave these in there, I know it's a bit of confusion but they *may* be useful in some scenarios we're not considering ( Apple does leave them in there for some reason or another ).

PR #118, opened 2015-06-24
Fixes for Swift 2.0 b2

bendjones:
Tests are passing locally, please let me know if I missed something. By the way I love Masonry and SnapKit!

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -400,7 +400,7 @@ private extension NSLayoutAttribute {
                 case .Leading: return point.x
                 case .Trailing: return point.x
                 case .Width, .Height, .NotAnAttribute: return CGFloat(0)
-                default: return CGFloat(0)
+                case .FirstBaseline: return point.y
                 }
             #endif
         }
@@ -425,7 +425,7 @@ private extension NSLayoutAttribute {
                 case .Leading: return  (Config.interfaceLayoutDirection == .LeftToRight) ? insets.left : -insets.right
                 case .Trailing: return  (Config.interfaceLayoutDirection == .LeftToRight) ? insets.right : -insets.left
                 case .Width, .Height, .NotAnAttribute: return CGFloat(0)
-                default: return CGFloat(0)
+                case .FirstBaseline: return insets.bottom
                 }
             #endif
         }
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -400,6 +400,7 @@ private extension NSLayoutAttribute {
                 case .Leading: return point.x
                 case .Trailing: return point.x
                 case .Width, .Height, .NotAnAttribute: return CGFloat(0)
+                default: return CGFloat(0)
                 }
             #endif
         }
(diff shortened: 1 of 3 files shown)


Source/Constraint.swift
```diff
@@ -424,6 +425,7 @@ private extension NSLayoutAttribute {
                 case .Leading: return  (Config.interfaceLayoutDirection == .LeftToRight) ? insets.left : -insets.right
                 case .Trailing: return  (Config.interfaceLayoutDirection == .LeftToRight) ? insets.right : -insets.left
                 case .Width, .Height, .NotAnAttribute: return CGFloat(0)
+                default: return CGFloat(0)
```
robertjpayne: Are you sure these are required? Seems like there may be new attributes we need to manage? We don't want to miss any new ones.

Source/Debugging.swift
```diff
@@ -167,6 +167,7 @@ private extension NSLayoutAttribute {
         case .CenterX:              return "centerX"
         case .CenterY:              return "centerY"
         case .Baseline:             return "baseline"
+        default:                    return "default"
```
robertjpayne: And again on this, I'd rather build in support for new cases now

robertjpayne:
@bendjones looks pretty good. Main thing is can we list out any new attributes with fatal errors rather than using a default catch all case? Just so we don't miss any and get weird/undefined behaviour happening.

bendjones:
Yeah let me add those missing cases to the tests quick. Totally agree @robertjpayne

bendjones:
Turns out the missing case is NSLayoutAttribute.FirstBaseline which the docs state as the following:

```
NSLayoutAttributeFirstBaseline
The object’s baseline. For objects with more than one line of text, this is the baseline for the topmost line of text.

Available in OS X v10.11 and later.
```

Not sure how we want to handle that in this case. Suggestions?

bendjones:
@robertjpayne I think the above commit handles that new case in a reasonable way. To be the use would be to offset by the Y or bottom in relation to the FirstBaseline. But if you have a different thought I'd love to hear it. I haven't used this constraint myself so I'm just basing my change on the docs.

Source/ConstraintAttributes.swift
```diff
@@ -157,10 +157,10 @@ internal struct ConstraintAttributes: RawOptionSetType, BooleanType {
     }
 }
 internal func +=(inout left: ConstraintAttributes, right: ConstraintAttributes) {
-    left = (left | right)
+    left = left.intersect(right)
```
jsslai: Should this be left.union(right)? Or replace with left.unionInPlace(right)

robertjpayne: yea I think this is wrong too -- it should be a union not a intersect @bendjones can you fix that?

bendjones:
Ok @jsslai and @robertjpayne fixed that intersect and changed them to inPlace operations.

robertjpayne:
LGTM thanks :+1: merged!

PR #121, opened 2015-06-30
Fix crash

alexanderjarvis:
This was causing an EXC Bad access crash.

### DIFF
--- a/Source/Debugging.swift
+++ b/Source/Debugging.swift
@@ -107,11 +107,11 @@ private var labelKey = ""
 private func descriptionForObject(object: AnyObject) -> String {
     let pointerDescription = NSString(format: "%p", [object])
     if let object = object as? View {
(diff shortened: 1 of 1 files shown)


robertjpayne:
@alexanderjarvis do you know if this was specific to Swift-2.0 or Swift 1.x?

alexanderjarvis:
Swift 1.2 (when a constraint was broken I think)

robertjpayne:
@alexanderjarvis awesome, looks good but please re-submit it against develop rather than master branch. Else I can manually commit it in?

alexanderjarvis:
@robertjpayne see #122

alexanderjarvis:
If you change the default branch to develop this will help with future PRs

robertjpayne:
@alexanderjarvis good point, I'll do that too and thanks for re-submitting this!

PR #127, opened 2015-07-12
Add source file location to debug description

aleffert:
Before:
```
<<SnapKit.LayoutConstraint:0x7f943483ea00> <UIView:0x7f9432ff2110>.width == 120.0>"
```

After:
```
<<SnapKit.LayoutConstraint:0x7f943483ea00@Tests.swift:30> <UIView:0x7f9432ff2110>.width == 120.0>"
```

The main code changes are:
- Rejiggering the ``ConstraintDescription`` hierarchy to be based on
  classes instead of protocols. This was necessary since protocols can't
  expose methods with default arguments which are used for picking up
  the location via ``__FILE__`` and ``__LINE__``.
- Propagating SourceLocation information from the
  ``ConstraintDescription`` to the ``ConcreteConstraint`` to the
  ``description`` method.
- Adding support for NSLayoutConstraint's ``identifier`` property

### DIFF
--- a/SnapKit.xcodeproj/project.pbxproj
+++ b/SnapKit.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		77F5D2191B530F3B00A1558A /* SourceLocation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 77F5D2181B530F3B00A1558A /* SourceLocation.swift */; };
+		77F5D21A1B530F3B00A1558A /* SourceLocation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 77F5D2181B530F3B00A1558A /* SourceLocation.swift */; };
 		EE4910981B19A26000A54F1F /* ViewController+SnapKit.swift in Sources */ = {isa = PBXBuildFile; fileRef = EE4910971B19A26000A54F1F /* ViewController+SnapKit.swift */; };
 		EE4910991B19A40200A54F1F /* SnapKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = EEBCC9D819CC627D0083B827 /* SnapKit.framework */; };
 		EE94F6091AC0F10A008767FF /* UIKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = EE94F6081AC0F10A008767FF /* UIKit.framework */; };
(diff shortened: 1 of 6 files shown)


robertjpayne:
@aleffert is there no way to accomplish this with protocol chains rather than class chains? It's one thing I'd really like to avoid especially with Swift's 2.0 protocol features on the horizon.

aleffert:
As best I can tell, ``ConstraintDescriptionPrioritizable`` needs to be a real class in order for the default arguments stuff to work and that would force ``ConstraintDescriptionEditable`` to be one as well. I don't believe Swift 2 changes this, but I'll check in a bit.

I think you could get away with using protocols for the other parts, though I haven't worked through the exercise. Would it be preferable to have some of the parts be classes and the rest protocols, or all classes for the uniformity?

Obviously, you're under no obligation to take this if you don't like either of those solutions, but if we can make it work, it seems like a nice improvement. In addition to the general debugging utility, I'm hoping to use the location information for a tooling project I'm messing around with (which would probably require it to be a public property on ``Constraint``, but that's independent of this PR).

aleffert:
Interestingly, it looks like Swift 2 will support this with a protocol based approach. (It looks like they do the resolution statically, which results in some surprising behavior, but does make this case work).

For example, the following builds successfully:
```
protocol ConstraintDescriptionPrioritizable {
    
}

extension ConstraintDescriptionPrioritizable {
    func updatePriority(priority : Float, file : String = __FILE__, line : UInt = __LINE__) {
    }
}
```

aleffert:
Thinking about it a bit more, I don't think you're going to get much out doing protocols that way. 

Since protocol methods with implementations are statically resolved, you don't get the polymorphism that using a protocol might otherwise imply. Also, since the implementation is on the protocol itself, you don't have access to any implementation details, so the protocol implementation would have to downcast itself to a ``ConcreteConstraint``, which, 1. is ugly, 2. doesn't support any other implementations of those protocols anyway.

robertjpayne:
@aleffert hmm yea it works but I'm also unsure on polluting the method namespace, I know that it's default values so the user doesn't have to fill them out but by default auto complete on my end is showing them and you have to manually delete them out. It's too bad we can't have "Hidden" default params to do something like this it really is a great feature to have

aleffert:
Huh. I'm not seeing it show up in autocomplete. I'd sort of assumed there was special handling, since it doesn't show up for auto completing XCTest things either.

aleffert:
The following example in a playground also doesn't have any autocomplete clutter
```

func test(file : String = __FILE__, line : UInt = __LINE__) {
    println("test \(file):\(line)")
}

```

autocompletes to just ``test()``

I'm on Xcode 6.4

robertjpayne:
@aleffert sorry I've not had time to dig into this and merge it up, I'm looking at next week for some availability!

aleffert:
Thanks for the update!

robertjpayne:
@aleffert so having a bit more time to review it, I'm still not entirely sold on it, would it be good enough to maybe track the file and line on the constraint maker itself rather than trying to proxy these files and line numbers all over?

aleffert:
I think that makes sense for most users and if you want to leave it there, that totally makes sense. My use case is a little more unusual. In particular, I have a desktop side debugging tool I wrote and I'm trying to add a feature where you can visually edit your constraints at runtime and those changes are automatically written back into your code. Having the individual line makes it a lot easier for me to find the right code to modify.

robertjpayne:
@aleffert ahh sounds very interesting would love to see it!

aleffert:
So maybe lets go with your simpler solution for now and if I manage to actually get around to making it work, you can decide then whether it's worth bringing in my more invasive changes? Happy to open a new PR at that point. We'll probably all be on Swift 2 by then anyway.

robertjpayne:
@aleffert sounds great! Thanks again for bringing this up though even the small support at pointing to the closure blocks makes a huge difference in debugging!

PR #137, opened 2015-08-27
Support for UILayoutSupport on iOS 9

rasmusth:
Layout guides are no longer UIViews on iOS 9 and the view variable on ConstraintItem fails to return. In those cases try the layoutSupport property on iOS before defaulting to installOnView.

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -241,9 +241,14 @@ internal class ConcreteConstraint: Constraint {
             let layoutConstant: CGFloat = layoutToAttribute.snp_constantForValue(self.constant)
             
             // get layout to
-            var layoutTo: View? = self.toItem.view
+            var layoutTo: AnyObject? = self.toItem.view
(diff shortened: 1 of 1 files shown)


robertjpayne:
Hey @rasmusth thanks for this PR! I'm slammed until this next weekend which hopefully Xcode 7 GM is out by then so I can push Swift 2.0 entirely live!

rasmusth:
Sure, take you time :)

robertjpayne:
Thanks for the PR, I'm patching it slightly different:

```
```
            #if os(iOS)
            var layoutTo: AnyObject? = self.toItem.view ?? self.toItem.layoutSupport
            #else
            var layoutTo: AnyObject? = self.toItem.view
            #endif
            if layoutTo == nil && layoutToAttribute != .Width && layoutToAttribute != .Height {
                layoutTo = installOnView
            }
```

rasmusth:
(Y)

robertjpayne:
This is now fixed on develop.

PR #139, opened 2015-09-04
Do not assume toItem is a UIView (Fix for aligning to layout guides in iOS 9)

sebreh:
On iOS 9 layout guides are not UIView subclasses anymore, but instead a private type `_UILayoutSpacer`, causing `snp_topLayoutGuideBottom` etc. to break because the toItem is no longer a `UIView`. There was a check in the code that would automatically pick another view if the toItem was not a `UIView`.

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -241,11 +241,8 @@ internal class ConcreteConstraint: Constraint {
             let layoutConstant: CGFloat = layoutToAttribute.snp_constantForValue(self.constant)
             
             // get layout to
-            var layoutTo: View? = self.toItem.view
-            if layoutTo == nil && layoutToAttribute != .Width && layoutToAttribute != .Height {
-                layoutTo = installOnView
-            }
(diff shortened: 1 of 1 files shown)


robertjpayne:
@sebreh sorry for delayed update, this looks good I'm going to do a little bit more testing with it first though.

sebreh:
:+1:

robertjpayne:
I think this is invalid logic, at least it reads as such. I'm instead patching it as follows:

```
            var layoutTo: AnyObject? = self.toItem.view ?? self.toItem.layoutSupport
            if layoutTo == nil && layoutToAttribute != .Width && layoutToAttribute != .Height {
                layoutTo = installOnView
            }
```

This ensures that layout guides are supported, but we still maintain specific rules for .Width and .Height attributes.

robertjpayne:
This is now fixed on develop.

sebreh:
Great, thanks!

PR #143, opened 2015-09-14
Add availability check for iOS 8 and OS X 10.10 APIs.

devxoul:
SnapKit's minimum target is set to 8.0, however, it is necessary for projects that support iOS 7.

e.g. Installing via [CocoaSeeds](https://github.com/devxoul/CocoaSeeds) or drag and drop.

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -138,24 +138,24 @@ internal class ConcreteConstraint: Constraint {
     }
     
     internal override func activate() -> Void {
-        if NSLayoutConstraint.respondsToSelector("activateConstraints:") && self.installInfo != nil {
-            let layoutConstraints = self.installInfo!.layoutConstraints.allObjects as! [LayoutConstraint]
-            if layoutConstraints.count > 0 {
-                NSLayoutConstraint.activateConstraints(layoutConstraints)
-            }
-        } else {
+        guard #available(iOS 8.0, OSX 10.10, *), self.installInfo != nil else {
             self.install()
+            return
+        }
+        let layoutConstraints = self.installInfo!.layoutConstraints.allObjects as! [LayoutConstraint]
+        if layoutConstraints.count > 0 {
+            NSLayoutConstraint.activateConstraints(layoutConstraints)
         }
     }
     
     internal override func deactivate() -> Void {
-        if NSLayoutConstraint.respondsToSelector("deactivateConstraints:") && self.installInfo != nil {
-            let layoutConstraints = self.installInfo!.layoutConstraints.allObjects as! [LayoutConstraint]
-            if layoutConstraints.count > 0 {
-                NSLayoutConstraint.deactivateConstraints(layoutConstraints)
-            }
-        } else {
-            self.uninstall()
+        guard #available(iOS 8.0, OSX 10.10, *), self.installInfo != nil else {
+            self.install()
+            return
+        }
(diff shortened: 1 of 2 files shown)


robertjpayne:
@devxoul am aware of this necessity but we also need to allow for OS X targeting 10.10 to use the activate/deactivate apis.

devxoul:
@robertjpayne, thank you for comment.
I've just updated the code and you'll love it.

- Add availability check for iOS 8 attributes
- Add availability check for iOS 8 and OS X 10.10 attributes: activating and deactivating constraints.

dennda:
Any chance we can get this merged? master doesn't currently build under Xcode 7. :-)

devxoul:
We can ping to @robertjpayne.

robertjpayne:
Master should build fine if you're targeting iOS 8.0 or OS X 10.10 but not iOS 7/10.9. I was away for a week and will be fixing soon.

dennda:
It doesn't build with the newly released Xcode 7 is what we were saying. I don't think it matters what you target.

FWIW I can confirm that these changes make it compile again.

Source/Constraint.swift
```diff
@@ -138,24 +138,32 @@ internal class ConcreteConstraint: Constraint {
...
-            let layoutConstraints = self.installInfo!.layoutConstraints.allObjects as! [LayoutConstraint]
-            if layoutConstraints.count > 0 {
-                NSLayoutConstraint.activateConstraints(layoutConstraints)
-            }
-        } else {
+        guard #available(iOS 8.0, OSX 10.10, *) else {
+            self.install()
+            return
+        }
+        guard self.installInfo != nil else {
```
dopcn: one guard statement can check two conditions separated by comma at the same time

devxoul: Awesome, thanks!

devxoul:
@dennda, alternatively I'm using [CocoaSeeds](https://github.com/devxoul/CocoaSeeds) instead of CocoaPods or something else for supporting iOS 7.

For example,

```ruby
github 'devxoul/SnapKit', 'availability', :files => 'Source/*.{swift,h}'
```

dennda:
@devxoul OK? I'm not sure what you're trying to tell me; I don't use CocoaPods/Seeds/Carthage. I already solved the problem locally, but it'd be nice if I could just go back to using master.

robertjpayne:
This has been merged into develop, I've also added availability checks to attributes that are only available on iOS 8.0+ so they can't get misused on lesser targets or OS X.

dopcn:
@robertjpayne Hi there. I'm confused by the commit right after you merge this pr. This pr makes Snapkit  can run on iOS7 but the commit after the merge reverse it. So am I right Snapkit will never support iOS7 officially?

robertjpayne:
@dopcn no it should still be fine? (though you must manually integrate SnapKit into iOS 7.0 targets.) The `@available` macros in the subsequent commit added ensure that you cannot use iOS 8.0 API's on iOS 7 deployment targets is all.

robertjpayne:
Ok `develop` now has fixes for this. Unfortunately I do require a build setting for anyone targeting legacy platforms. You can read more about that here: https://github.com/SnapKit/SnapKit/blob/develop/LEGACY_PLATFORMS.md

dennda:
404 sir.

robertjpayne:
@dennda thanks! Sorry I had updated it and moved it to the main website after I posted that comment!

PR #151, opened 2015-09-23
Add support for the iOS 9 Layout Anchor API.

aleffert:


robertjpayne:
@aleffert thanks for this :) Was meaning to get around to these but figured nobody will be using them for a bit anyways!

aleffert:
I had some weird bugs with layout guides on view controllers that were solved by switching to the analogous layout anchors.

PR #153, opened 2015-09-25
Added tvOS checks so the iOS paths are used rather than OS X

brittonmobile:
This adds a check for tvOS so that it uses the same code paths as iOS.

I did not add a tvos target to the podspec yet because that Cocoapods branch is still unreleased. Until then, this workaround in a Podfile is sufficient to get it up and running on tvOS:

	post_install do |installer|
			installer.pods_project.targets.each do |target|
					target.build_configurations.each do |config|
							config.build_settings['SUPPORTED_PLATFORMS'] = ['appletvos', 'appletvsimulator']
					end
			end
	end

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -21,7 +21,7 @@
 //  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 //  THE SOFTWARE.
 
-#if os(iOS)
+#if os(iOS) || os(tvOS)
 import UIKit
 #else
 import AppKit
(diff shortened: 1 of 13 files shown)


robertjpayne:
@brittonmobile what about the `@available` macros. I assume we have to update those for tvOS? Or maybe since there is the `*` and it's a newer platform it's all supported?

brittonmobile:
The `*` in the `@available` macros seems to cover it all. Since tvOS 9.0 is the first version, there's no older versions to worry about.

robertjpayne:
@brittonmobile thanks again, merging this.

PR #163, opened 2015-10-15
Add source location information to each constraint

aleffert:
I got my tooling working so I'm back with this change. You can check it out here: https://github.com/aleffert/snaps or just watch a video https://aleffert.github.io/snaps/Documentation/snaps-example.html to see some live constraint editing + code updating.

This version has some differences and improvements from the last PR. I'm not super attached to many of the particular details of this rather large change so long as line information ends up available, but I do think the only way to make that work is switch from protocols to classes. As best I can tell, the line information does not clutter up autocomplete. I also factored things a bit so autocomplete doesn't have to have Int/Float/UInt/CGFloat/etc. for each function and generally makes it easier to add new types. But if you preferred having each type broken out, I'm happy to add the extra code.

Also, since this a big change, I broke it up into a number of more digestible commits. If you want me to do separate PRs with only some of it or whatever I'm happy to do that.

### DIFF
--- a/SnapKit.xcodeproj/project.pbxproj
+++ b/SnapKit.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		779230441BCB6FC30027CF5F /* SourceLocation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 779230431BCB6FC30027CF5F /* SourceLocation.swift */; settings = {ASSET_TAGS = (); }; };
+		779230451BCB6FC30027CF5F /* SourceLocation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 779230431BCB6FC30027CF5F /* SourceLocation.swift */; settings = {ASSET_TAGS = (); }; };
 		EE4910981B19A26000A54F1F /* ViewController+SnapKit.swift in Sources */ = {isa = PBXBuildFile; fileRef = EE4910971B19A26000A54F1F /* ViewController+SnapKit.swift */; };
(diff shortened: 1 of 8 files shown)


robertjpayne:
@aleffert awesome! I'll review it in detail over the weekend

robertjpayne:
@aleffert I REALLY like this, I may make some more tweaks but pulling this into develop right now. Thanks so much.

aleffert:
Awesome. Are you likely to cut a release soon? I'd like to be able to point my code at a real version.

robertjpayne:
@aleffert I had a family member pass away suddenly so I'm traveling and unable to really move it along until next week. Otherwise was hoping to get it released this last weekend.

aleffert:
My condolences. Obviously, that's way more important.

robertjpayne:
@aleffert this has been pushed to trunk now :+1: thanks again!

aleffert:
Woo. Thanks

robertjpayne:
@aleffert I'm going to have to revert this for the time being until #169 is resolved. I think it's a minor fix but I'm worried about other side effects.

Reality is I should probably have more tests that cover all the use cases so that big changes like this don't break anything but I don't have those in place yet.

I also missed that we've converted everything to a class and are allocating a lot of objects during the DSL chain which was something I was hoping we could avoid ( though maybe it's not plausible ) and missed when merging this in.

aleffert:
No worries. Let me know if there's a way I can help. I'd be surprised if the allocations were a big deal - they're pretty small compared to a bunch of stuff that typically happens during layout, but since the wrappers are pretty dumb, it'd be easy to do a pool allocator or something if that ends up actually being a problem.

robertjpayne:
@aleffert so unfortunately I'm going to have to indefinitely let this PR rest in peace. It has caused a bit of headache around composite constraints and I didn't have time to suss out exactly what was happening.

Regardless I've been working on a major revamp of SnapKit to help make the source easier to follow and maintain. You can track progress on #191 which includes per-constraint source information.

PR #175, opened 2015-12-11
Create Package.swift

gitpark:
Package.swift


robertjpayne:
@gitpark can you update the license to match other source files in the project?

PR #180, opened 2016-01-05
Added Sample Demo application.

neelamsharma12:


robertjpayne:
@neelamsharma12 I appreciate this PR! I'll try to squeeze some time in to review and merge :+1:

robertjpayne:
@neelamsharma12 unfortunately I cannot accept this PR as it contains unecessary pieces:

* You are using Cocoapods to install the dependency where as it would be better to use SnapKit via a Subproject.
* You have not added the Pods folder to .gitignore so the entirety of the Pods folder is now committed
* There is a CoreData stack that is being created
* Preference would be to not include any Storyboard files as they are not necessary to illustrate SnapKit's functionality
* Preference would be to not include any image files as they are not necessary to illustrate SnapKit's functionality

I'm going to close this PR for now. I do appreciate the effort so if you do have time still and would like to fix the above issues and re-submit a new PR please do so.

PR #189, opened 2016-01-21
Update CHANGELOG.md

StefMa:
I think we can not actually back to the future ;)

Changed release date from 0.19 from **2015** to **2016**


robertjpayne:
Whoops thanks!

PR #191, opened 2016-01-26
SnapKit - 0.30.0

robertjpayne:
A long preview of what is to come. Currently mostly functional but not quite ready to go.

PR #193, opened 2016-01-29
Size classes support

nakiostudio:
First things first: congrats to the SnapKit team and contributors, this library simply makes constraints something pleasant to deal with.

This PR is a proof of concept of something that I find quite useful and which have let me save a lot of time and code. The change basically checks whether `installOnView` has the size classes for the given `Constraint` and, if not, the `NSLayoutConstraint` is not created.

I haven't had the time to study the source code in deep and I'm sure there must be a way of implementing this better and, more important, according with the general architecture of the library.

Find below some sample code:

```
    override func traitCollectionDidChange(previousTraitCollection: UITraitCollection?) {
        super.traitCollectionDidChange(previousTraitCollection)
        
        self.searchField.snp_updateConstraints { make in
            make.width.lessThanOrEqualTo(400).priorityLow()
            make.top.bottom.equalTo(self).inset(5)
            make.right.left.equalTo(self).inset(5).forHorizontalSizeClasses([.Unspecified, .Compact])
            make.centerX.equalTo(self).forHorizontalSizeClasses([.Regular])
        }
    }
```

This will basically center the `searchField` horizontally on large screen sizes whereas the `searchField` will fill the superview edge to edge on small screen sizes.

Thanks in advance for the feedback and give it a try!


mattvagni:
:+1:

Source/Constraint.swift
```diff
@@ -61,6 +61,11 @@ public class Constraint {
     internal var makerFile: String = "Unknown"
     internal var makerLine: UInt = 0
     
+    public var layoutConstraint: LayoutConstraint? {
```
robertjpayne: This isn't a valid API because a Constraint can represent more than 1 LayoutConstraint and by not exposing it we can ensure SnapKit always has total ownership over the LayoutConstraints.

robertjpayne:
@nakiostudio I like the idea but the implementation is a bit flawed:

1. The way the API reads it would seem that the constraints will automatically activate/deactivate when the trait collection changes. Where as in reality the user needs to manually update/remake them.

2. Your example will break because snp_updateConstraints wouldn't really be able to support this. While for now you've not added the size classes to the equality set it'd actually mean that they were never taken into account and you would just update the existing size classes constraint constant. Vice versa if we add the size classes into the equality chain it wont allow you to update based on size class shifts.

Personally if we can't make them magically activate/deactivate I see little value in adding it into the constraint DSL. It'd be better off as an extension on UITraitCollection that allows less verbose API for checking/switching.

robertjpayne:
We have a discussion over on Masonry about this and an idea for breakpoints https://github.com/SnapKit/Masonry/pull/293. The idea is it would be *amazing* to be able to install Constraints that automatically update the underlying NSLayoutConstraints based on things like trait collection, device etc…

Source/Constraint.swift
nakiostudio: Shame on me! I did expose this on my fork to allow some constraint animations with Pop, I should have excluded this from the PR >_<'

nakiostudio:
The constraints automatically enabled/disabled sounds ideal indeed, I will try a different approach and take a look at the Masonry PR. 

Thanks for the comments @robertjpayne :)

nakiostudio:
Hi @robertjpayne!
I've made some changes in my code and now the constraints are automatically updated with no need of overriding the `traitCollectionDidChange:` method by the client apps. However, a recreation of the constraints takes place under the hood as `NSLayoutConstraints` don't have a real deactivated state. Hope you like the approach, I think we are getting closer!

Panajev:
Amazing effort :)!

nakiostudio:
I've updated its usage to make it a bit cleaner, now the sample implementation would look like this:
```
self.searchField.snp_updateConstraints { make in
       make.width.lessThanOrEqualTo(400).priorityLow()
       make.top.bottom.equalTo(self).inset(5)
       make.right.left.equalTo(self).inset(5).horizontalSizeClass(.Compact)
       make.centerX.equalTo(self).horizontalSizeClass(.Regular)
}
``

File: Source/SizeClass.swift [hunk 2]
```diff
@@ -0,0 +1,49 @@
...
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+import UIKit
```
robertjpayne: This will break OS X build targets.

Source/View+SnapKit.swift
```diff
@@ -178,6 +178,58 @@ public extension View {
             objc_setAssociatedObject(self, &installedLayoutConstraintsKey, newValue, objc_AssociationPolicy.OBJC_ASSOCIATION_RETAIN_NONATOMIC)
         }
     }
+    
+    internal var snp_constraints: [Constraint] {
```
robertjpayne: We shouldn't need this property on the view at all. There is a snp_layoutConstraints which you can map back into the originally Constraints. Was that not suitable?

Source/View+SnapKit.swift
```diff
@@ -178,6 +178,58 @@ public extension View {
...
+        set {
+            objc_setAssociatedObject(self, &constraintsKey, newValue, objc_AssociationPolicy.OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+        }
+    }
 }
 
 private var installedLayoutConstraintsKey = ""
+private var constraintsKey = ""
+
+public extension UIView {
```
robertjpayne: We need to get rid of this swizzle, it's not worth the risk. Probably going to involve a hidden subview we insert and track and remove when necessary.

robertjpayne:
@nakiostudio there's still a few major issues:

1. Make sure the OS X build target works. Pretty sure we have some rogue imports and logic that would cause it to fail.

2. Need to find a way to do this without a swizzle. I'm just not willing to deploy a swizzle to a library thousands of apps currently depend on. It's fine for individuals that would want to integrate it themselves and accept that risk but not a decision we should make for individuals.

I do apologize too I know the codebase isn't the easiest to work with. There's a huge refactor already happening for #191.

Source/SizeClass.swift
nakiostudio: True, updated!

Source/View+SnapKit.swift
nakiostudio: I'm afraid `snp_layoutConstraints` doesn't work in this case because we need to store the `Constraint` objects that will create the `LayoutConstraints` for some size classes. If the `LayoutConstraint` doesn't exist yet then there is no way we can reverse it and obtain our `Constraint`. Also we need these constraints stored in `toInstallView` instead of `fromView`.

nakiostudio:
@robertjpayne Thanks for your feedback. 

I've fixed point 1 and about point 2... I think you are right, method swizzling is always risky but at the same time it is the only way we can detect the traitCollection changes for the view. We could sort this out by providing a method the user can call from `traitCollectionDidChange:`

Don't apologize for the codebase, I've had a great read!

nakiostudio:
I think size classes + automatic enabling/disabling would be a game changing feature for library, I'm going to give a try to that hidden UIView idea.

nakiostudio:
@robertjpayne I've implemented your hidden view idea and it does work. Now my concern is what would devs think about that ghost UIView swarming around.

robertjpayne:
@nakiostudio far better they are aware we have to do some tricky stuff than having a swizzle…

nakiostudio:
True :) It's midnight here in London but tomorrow I will also ensure that this view is not created unless size classes different than `.Any` are defined.

nakiostudio:
@robertjpayne I've made a few changes. Avoided some code duplication and ensured that no more than one `ListenerView` is created per `UIView`.

nakiostudio:
Any feedback on this @robertjpayne? :)

Panajev:
Was it merged? Why did you close it?

[[iOS messageWithContent:webContent] broadcast]

nakiostudio:
Hi @Panajev :wave: 
I don't think there is much interest on getting this merged so I've decided to keep it in my fork along with other changes I have in mind :)

Panajev:
This would be quite sad, I hope we can change the maintainer's mind. You addressed the concerns. Please reopen it :).

[[iOS messageWithContent:webContent] broadcast]

nakiostudio:
Hi @Panajev :wave: 
If you are still interested in the feature give a try to this https://github.com/nakiostudio/EasyPeasy. It's the first version and is not very documented yet but little by little :rocket: The example project has plenty of sample code though :+1:

Panajev:
Thanks, I will take it a look :).

[[iOS messageWithContent:webContent] broadcast]

Issue #210, opened 2016-03-21
[FEATURE] Add constraint identifiers as an option during ConstraintMaker finalization.


rohan-panchal:
### New Issue Checklist

- [X] I have looked at the [Documentation](http://snapkit.io/docs)
- [X] I have read the [F.A.Q.](http://snapkit.io/faq)

##### Platform

* [X] OS X
* [X] iOS

##### Versions

* Platform version is: `iOS 9.0`
* SnapKit version is: `0.19.1`

##### Integration

- [ ] Carthage
- [X] Cocoapods
- [ ] Manually / Other

### Issue

Having identifiers on constraints allows for easier debugging when constraints are broken:

http://useyourloaf.com/blog/using-identifiers-to-debug-autolayout/

I was hoping for a solution to add identifiers to each constraint. as part of the ConstraintMaker syntax.

My thought was to add a method:

    `func identifiedAs(string: String) -> ConstraintDescriptionFinalizable`

to `ConstraintDescriptionFinalizable`

as well as a variable to `ConstraintDescription` leaving it optional, and adding the actual setting of the NSLayoutConstraint property in `installOnView`.

If thats ok with you i can make a pull request as soon as I'm done testing it.

robertjpayne:
:+1: Sounds good to me!

PR #211, opened 2016-03-21
Added identifier field to ConstraintDescription

rohan-panchal:
Adds identifiers to LayoutConstraints which therefore get added to the NSLayoutConstraint.

References issue: https://github.com/SnapKit/SnapKit/issues/210

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -193,15 +193,18 @@ internal class ConcreteConstraint: Constraint {
         }
     }
     
+    private let identifier: String?
+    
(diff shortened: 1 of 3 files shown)


robertjpayne:
@rohan-panchal sorry just reading through the commit and only thing any chance we can make the API just `labeled(XXX)`?

Just think it'll read better and make more sense to end users?

rohan-panchal:
Agreed on the labeled method.

Updated.

robertjpayne:
Thanks! I'll probably do a release in about 4-5 days with it :+1:

PR #216, opened 2016-03-25
Removed Prefixed APIs

MP0w:
Swift should use static dispatch for methods, properties and functions that use features not available in objc, instead for categories of objc types and methods that are @objc (will use dynamic dispatch) we can use `@objc(snp_method) func method()` to make sure we don't override private APIs

Would you consider this Breaking change? It looks like useless to use prefixes in swift extension [except  some cases](https://pspdfkit.com/blog/2016/surprises-with-swift-extensions/) where @objc(prefix_name) can be used

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -178,7 +178,7 @@ internal class ConcreteConstraint: Constraint {
             if let installInfo = self.installInfo {
                 for layoutConstraint in installInfo.layoutConstraints.allObjects as! [LayoutConstraint] {
                     let attribute = (layoutConstraint.secondAttribute == .NotAnAttribute) ? layoutConstraint.firstAttribute : layoutConstraint.secondAttribute
-                    layoutConstraint.constant = attribute.snp_constantForValue(self.constant)
+                    layoutConstraint.constant = attribute.constantForValue(self.constant)
                 }
             }
         }
@@ -251,7 +251,7 @@ internal class ConcreteConstraint: Constraint {
             let layoutToAttribute = (layoutToAttributes.count > 0) ? layoutToAttributes[0] : layoutFromAttribute
             
             // get layout constant
-            let layoutConstant: CGFloat = layoutToAttribute.snp_constantForValue(self.constant)
+            let layoutConstant: CGFloat = layoutToAttribute.constantForValue(self.constant)
             
             // get layout to
             #if os(iOS) || os(tvOS)
@@ -278,7 +278,7 @@ internal class ConcreteConstraint: Constraint {
             layoutConstraint.priority = self.priority
             
             // set constraint
-            layoutConstraint.snp_constraint = self
+            layoutConstraint.constraint = self
             
             newLayoutConstraints.append(layoutConstraint)
         }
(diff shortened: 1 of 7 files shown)


robertjpayne:
@MP0w I wonder if I can somehow build this into a subspec and compiler flag. I worry about removing the `snp_` prefixes for some of the variables like `snp_height` as `height` may be a common extension?

Otherwise this looks pretty good, helps clean up the library a lot.

MP0w:
I did some tests and tried to understand all the edge cases. Looks like we don't need any prefix to prevent private APIs collision.

Scenarios:
1. Swift extension of non-objc type -> No prefix needed, static dispatch is used so can't override private APIs
2. Swift extension of objc type with method available to objc (doesn't use swift only features) -> Use dynamic dispatch so has to be prefixed. But using `@objc(sip_methodName)` the method will be prefixed in objc and you can safely use it without prefix in swift without private APIs collision (namespace inferred and other modules could use the same name)
3.  Swift extension of objc type with method unavailable to objc (most common in SnapKit) -> Use static dispatch, not exposed to objc and won't interfere with private APIs

I tried adding an objc category that defines `height` in a separate dynamic framework (to simulate UIKit?)

```
    @objc(snp_height) var height: String { return "swift extension" } // needs @objc(rt_height) because is exposed to objc
    var width: SomeStruct { return SomeStruct() } // no need for @objc() because struct can't be used anyway and the API is not exposed to objc
```

From swift we only see `height` and `width` (and if we import the category in the bridging header the compiler complains, that would be what happens if the SDK add a public API with the same name, same problem would happen for subclasses)

From objc we see:
```
NSLog(@"%@", @(self.view.height)); // print the view's height
NSLog(@"%@", self.view.snp_height); // print "swift extension"
// width can't be used at all.
```

robertjpayne:
@MP0w there is potentially high collision risk with someone that has this in their apps code base isn't there?

```
// Existing
extension UIView {
  var height: CGFloat { return self.frame.size.height }
}
```

I'll need to just double check that use case and see what Xcode does.

MP0w:
@robertjpayne true, I just checked and:
```
extension UIView {
    @nonobjc var test: Int { return 1 }
    @nonobjc var test: String { return "" }
}
```
is an invalid redeclaration. Instead:
```
extension UIView {
    @nonobjc func test() -> Int { return 1 }
    @nonobjc func test() -> String { return "" }
}
```
Is valid.

MP0w:
I will ask clarifications in the swift evolution, I'm not understanding why two methods with the same name are non ambiguous but 2 properties or 1 method and 1 property are ambiguous.
```
extension Array {
    var test: Int { return 1 }

    func test() -> String { return "" }
}
```

I'm also wondering if then it would be cleaner to move the risky-properties somewhere else and keep what we can un-prefixed, like:
```
view.snp.height
```

robertjpayne:
@MP0w Yea it's tricky, I'm not keen to split the API into half prefixed and half unprefixed. My work in the `feature/0.30.0` branch is already moving to `.snp.` rather than `.snp_` prefixing on all variables and methods.

MP0w:
#191 looks cool. Will close this PR

robertjpayne:
@MP0w still have a bit of work to do on 0.30.0! But yea argh it would be so nice to get rid of the prefixes!

PR #226, opened 2016-04-19
Added convenient super view methods to ConstraintDescriptionRelatable

jlalvarez18:
This feature would be wonderful if added. You no longer need to keep track of what the superview is and no longer need to `make.leading.equalTo(self.superview!)`. Now all you would need to do is `make.leading.equalToSuperview()` There is also a fatal error thrown if the view has no superview.


robertjpayne:
@jlalvarez18 this is a pretty good idea, at first glance I was thinking it wasn't great to add convenience just for the sake of it but then realised we already do that for `priorityHigh()` etc…!

jlalvarez18:
@robertjpayne thanks! When are you planning on merging the latest PRs?

PR #227, opened 2016-04-19
Enabling app extension API only for tvOS and OSX targets

JamieWhite:
This is already done for iOS.


robertjpayne:
@JamieWhite thanks :)

Issue #158, opened 2015-10-04
Add another action to .travis.yml to also build and test with OS X target.


TadeasKriz:
This would help to make sure no PR will break the codebase, just because the author did not try building for OS X, which is something anyone can easily forget to do and CI should make sure it does not happen.

money150604:
Legacy Platforms (iOS 7.0, OS X 10.9)  ——> s.ios.deployment_target = '8.0'
不是说支持7.0嘛？。。

PR #228, opened 2016-04-26
Fixes tests not compiling on the OS X target and adds a line for the travis.yml so that tests are always ran

vittoriom:
This should also address #158


robertjpayne:
@vittoriom thanks for this super appreciate!

PR #236, opened 2016-05-19
SnapKit 0.30.0

robertjpayne:
**SnapKit 0.30.0** is a complete re-write to take advantage of modern Swift as well as
clean up the API. As such please use with caution and check your apps thoroughly

* All APIs are now accessed via `.snp.*` rather than `.snp_*`. For example you now do `view.snp.makeConstraints {…}`
* Debugging output has been improved to show exact file/line number from which the broken constraint was created. This is an improvement on existing functionality that showed only the file/line number for the make closure.
* `.inset()` will no longer take `CGPoint` or `CGSize` as input
* `.offset()` will no longer take `CGPoint`, `CGSize`, `EdgeInsets` as input
* Exceptions thrown during constraint creation have been replaced with `fatalError` with better output.


geoffmacd:
.snp.* is a curious choice. Care to explain the need for this? Do you view this a more idiomatic API?

robertjpayne:
@geoffmacd three things:

1. The API has to be prefixed because it's using UIKit extensions (it almost doesn't need to be, functions no longer need to be prefixed with Swift in this manner to avoid collisions but variables do).

2. I personally despise the underscore, it's hard a harder gesture to do and the SourceKit at the moment fairs way better with auto complete when the scope is limited.

3. It's much easier to break down the API when everything is done through `snp.` no longer do we have to write a class extension for each method on each UIKit class rather there is one entry point in any class extension to SnapKit related APIs.


I originally saw this proposal for RxSwift rather than having `rx_` on everything to do a `rx.` syntax. Because Swift structs are inexpensive to create and full featured theres practically no penalty performance wise.

If too many people view this as a big departure from `snp_` I'll probably leave both API's enabled or use a Cocoapods Subspec to enable the legacy API.

Bogidon:
@robertjpayne I like `.snp`. Easy to refactor to with a simple find and replace.
👍

adrianbukros:
Is it possible that the `removeConstraints` function is not working because the implementation is missing in 0.30.0.beta1? I checked the `ConstraintMaker` class and the function seems empty.

rpranata:
@robertjpayne Are you planning to also incorporate support for Swift 3.0 in this branch?

robertjpayne:
For anyone following this see 0.40.0 for Swift 3.0 support.

robertjpayne:
I'm going to close this PR, you can still use 0.30.0 branch but for now the API will be migrated with Swift 3.0

Issue #241, opened 2016-05-31
Getting at `LayoutConstraint`


startupthekid:
### New Issue Checklist

* [x] I have looked at the [Documentation](http://snapkit.io/docs)
* [x] I have read the [F.A.Q.](http://snapkit.io/faq)

### Issue Info

 Info                    | Value                               |
-------------------------|-------------------------------------|
 Platform                | e.g. ios
 Platform Version        | e.g. 9.0
 SnapKit Version         | e.g. 0.21.0
 Integration Method      | e.g. cocoapods
 

### Issue Description

Snapkit has a public class, `LayoutConstraint` which inherits from `NSLayoutConstraint` but there's no way to actually get an instance of `LayoutConstraint`. All the methods return `Constraint`. This becomes a problem when using libraries like Facebook's `pop` animation library because animations on constraints are applied directly to the constraint, as `constraint.pop_addAnimation` where constraint is a subclass of `NSLayoutConstraint`.

How can I get at `LayoutConstraint`?

robertjpayne:
@startupthekid this is intentional because SnapKit constraints are rarely a 1:1 relation to a LayoutConstraint.

In these cases I recommend people use the standard iOS API's to create the small subset of constraints needed to use with other frameworks/libraries.

colinta:
Is there any chance this can have more discussion?  You say "rarely a 1:1 relation", but I'm not sure this is true, it seems like SnapKit's DSL encourages the `top/left/right/bottom` constraints, which DO lend themselves to this "native access".  The other cases are center & size (1:2) and edges (1:4), which are easy to spot.

I would ask this: is there ever a time when you *might* get 1 constraint, or *might* get >1 constraints back? e.g. aren't the `make.foo` helpers always deterministic WRT how many constraints are created?

My use case is that I need the `NSLayoutConstraint` object that is created by `make.top.equalTo(view)` (to show/hide the nav bar via slide in/out animation).  Ideally I would accomplish this by:

```
navigationBar.snp_makeConstraints { make in
  make.left.equalTo(view)
  make.right.equalTo(view)
  let c = make.top.equalTo(view)
  self.navigationBarTopConstraint = c.nativeConstraints.first!
```

What you're recommending looks something like this, if I understand correctly:

```
navigationBar.snp_makeConstraints { make in
  make.left.equalTo(view)
  make.right.equalTo(view)
  let c = NSLayoutConstraint(bah,
    this: darn,
    mess: which,
    goes: on,
[...]

PR #242, opened 2016-06-14
Exposes "native constraints"

colinta:
This is a minimally invasive attempt to expose the `LayoutConstraint` objects that SnapKit creates, and is a solve for #241.

When `nativeConstraints()` is called on `ConcreteConstraint`, it first checks to see if the constraint(s) have been installed, and calls `install()` if `installInfo == nil`.  I checked this for sanity by looking at how often `install()` is called already, and it looked like it was written in such a way that it can be called more than once; in the spec code, it *is* called more than once on many of the constraints.  So hopefully this is an 👌 thing to do.

The spec code just does a few sanity checks: how many constraints are created, and do they have the expected `constant` values.

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -58,7 +58,7 @@ public class Constraint {
     public func updatePriorityMedium() -> Void { fatalError("Must be implemented by Concrete subclass.") }
     public func updatePriorityLow() -> Void { fatalError("Must be implemented by Concrete subclass.") }
 
-    public func nativeConstraints() -> [LayoutConstraint] { fatalError("Must be implemented by Concrete subclass.") }
+    public var layoutConstraints: [LayoutConstraint] { fatalError("Must be implemented by Concrete subclass.") }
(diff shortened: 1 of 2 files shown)


colinta:
Something that just occurred to me, is that the order of `[LayoutConstraint]` is not at all defined.  I wonder if:

(a) do we even care; the most common use case is `nativeConstraints().first`
(b) another "do we care?" point: expect the user to iterate through the array and filter out the constraint they want.
(c) would a lookup function be more appropriate, e.g. `nativeConstraints(.Top) -> LayoutConstraint?`
(d) or a dict? e.g. `nativeConstraints()[.Top]`

colinta:
Great!  Thanks for checking it out, and no worries if it turns out to be headache and needs to be pulled. Updated to use `var layoutConstraints`

robertjpayne:
@colinta we shall see, I think most people never need them so hopefully it's a non-issue.

PR #243, opened 2016-06-15
Feature/0.40.0 - Swift 3.0

robertjpayne:


knellr:
This is great and it's working nicely in my project 👍 

However - we could really do with adding `@discardableResult` to the phrase-ending function calls (equalTo, offset, inset etc). Without it there are warnings everywhere that the result is unused :s

bjarkehs:
Did I miss something or was updateOffset just straight up removed with no replacement?

robertjpayne:
@bjarkehs 0.40.0 was a full re-write to take advantage of protocols, clean up the API and sorts. As such I've not yet patched this back in.

robertjpayne:
@mattsrobot in tag 0.40.0.beta3 `center` should now work appropriately
@bjarkehs in tag 0.40.0.beta3 there is now API's for updating constraints

bjarkehs:
@robertjpayne sounds good. Luckily I was able to get the minimum deployment target bumped from iOS 7 to iOS 8, so I could continue to use SnapKit. It doesn't really make much sense to support iOS 7 when supporting iOS 10 anyway.

robertjpayne:
@bjarkehs yeah, to be honest too I've had way to many support requests about how to integrate with iOS 7.0 because lack of dynamic frameworks / static libs. It's been much much more hassle than it's worth and if people really still have to go that low they can always fork and modify a tiny bit that will be required!

bjarkehs:
@robertjpayne Or you know... Use Masonry. Which was what I ended up doing before convincing project manager that iOS 8 should be new deployment target. Anyway, keep up the good work! 👍

madhavajay:
@robertjpayne great job. Only thing I noticed is the testSizeConstraints test fails "sometimes". I ran it first time with the iPhone SE simulator selected and saw the issue. Subsequent runs pass or fail depending on something random? Not sure what could be causing it? Race condition with rounding issue perhaps?

screen shot 2016-08-31 at 09 20 21

robertjpayne:
@madhavajay yea it fails because the ordering of the constraints isn't guaranteed and I've been a little lazy fixing it.

Basically NSLayoutConstraint.activateConstraints manages the actual final installation of the constraints now and SnapKit has no control over the order in which they are added.

Fogh:
Has `inset` changed the way it works in this release? I have to use negative values for `offset` instead to get the same behavior. Is that intended?

robertjpayne:
@Fogh yup! I understand it may break a bit of code but I'm preferring to migrate along a major Swift version upgrade that is going to result in extra QA across the board.

You can track the changelog here: `https://github.com/SnapKit/SnapKit/blob/feature/0.40.0/CHANGELOG.md`

Fogh:
@robertjpayne So is `inset` going to be removed from the API?

robertjpayne:
@Fogh no it is not going to be removed, it was tweaked so that you naturally provide positive values for bottom/right is all.

PR #245, opened 2016-06-16
Add support for Swift 2.3

ketzusaka:
This updates the code to work under both 2.2 and 2.3 while developers handle their transition period up to Swift 3. The project file is also updated to indicate the last migration, and the version of swift. The version instructs Xcode 8 which internal toolchain is to be used.

### DIFF
--- a/SnapKit.xcodeproj/project.pbxproj
+++ b/SnapKit.xcodeproj/project.pbxproj
@@ -463,9 +463,11 @@
 					};
 					EEBCC9D719CC627D0083B827 = {
 						CreatedOnToolsVersion = 6.0;
+						LastSwiftMigration = 0800;
 					};
 					EEBCC9E119CC627D0083B827 = {
 						CreatedOnToolsVersion = 6.0;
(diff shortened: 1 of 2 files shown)


robertjpayne:
@ketzusaka any reason for the baseline change? Was Xcode giving a compiler error on this?

ketzusaka:
Indeed, that was the onl compile error.

robertjpayne:
@ketzusaka ok I'll check it out in a bit, the API diffs don't show any change to that so I'm curious as to what is erring because the change you've made isn't necessarily right.

ketzusaka:
Sounds good! This was what the Xcode migratory suggested; they may of gotten it wrong though :)

nickkjordan:
I believe this is correct – it seem like `Baseline` is no longer included in Xcode 8.

robertjpayne:
@nickkjordan this is tricky, because Baseline is supported down to iOS 6 where as LastBaseline is only available in iOS 8 yet we support iOS 7…

PR #253, opened 2016-07-03
Make the ConstrainMaker view property public

nickdomenicali:
Making the view public would allow common layout patterns to be reused.

```
func snapToEdges(make: ConstraintMaker) {
    guard let superview = make.view.superview else { return }
    make.edges.equalTo(superview)
}
numberLabel.snp_makeConstraints(closure: snapToEdges)
otherLabel.snp_makeConstraints(closure: snapToEdges)
```

### DIFF
--- a/Source/ConstraintMaker.swift
+++ b/Source/ConstraintMaker.swift
@@ -126,7 +126,7 @@ public class ConstraintMaker {
     
     internal let file: String
     internal let line: UInt
-    internal let view: View
+    public let view: View
     internal var constraintDescriptions = [ConstraintDescription]()
     
     internal func makeConstraintDescription(attributes: ConstraintAttributes) -> ConstraintDescription {


igorefremov:
@nickdomenicali @robertjpayne 

From my perspective, this decreases the overall readability, which seems to be one of the bigger goals of SnapKit.

i.e. 
```
//This is very clear as to what is going on
make.left.equalTo(self.someView)

//Eventually, it becomes clear as to what's happening here
//but, make.view is nowhere near as clear as make.left, make.right,
//make.edges, etc.
make.left.equalTo(make.view.superview ?? self.someView)
```

Couldn't you do something like this to achieve what you're looking for:
```
func snapToEdges(view: UIView) {
    guard let superview = view.superview else { return }
    view.snp_makeConstraints { (make) in
        make.edges.equalTo(superview)
    }
}

snapToEdges(numberLabel)
snapToEdges(otherLabel)
```

However, this PR did give me an idea, as I snap edges to superviews quite a bit.  I have just submitted https://github.com/SnapKit/SnapKit/pull/279 to make this faster.

robertjpayne:
@nickdomenicali I agree with @igorefremov here. It reduces readability a lot at the sake of being able to copy and paste constraint blocks.

Code should always be optimised for readability as it's always read far more often than it's written and constraint code is already too hard to figure out!

Do appreciate the contribution request though, thanks you!

PR #259, opened 2016-07-18
Fix various issues with legacy support in 0.40.0

bjarkehs:
Working on a project that has iOS 7 as deployment target. This seems to be the way to say that something is only available for iOS 8. I didn't have an iOS 8 project immediately available but if you wrap access to these properties in the following, then it still says deprecated. 

```
if #available(iOS 8.0, *) {

}

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -220,14 +220,14 @@ public class Constraint {
         }
         
         guard let installedLayoutConstraints = self.installInfo?.layoutConstraints.allObjects as? [LayoutConstraint]
-              where installedLayoutConstraints.count > 0 else {
-            return
+            where installedLayoutConstraints.count > 0 else {
+                return
         }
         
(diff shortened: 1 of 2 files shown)


robertjpayne:
@bjarkehs thanks for this I'm still on leave for two more weeks so may be delayed pulling this in, will try to in a few days when I have better connectivity to review!

robertjpayne:
@bjarkehs thanks again for this, temporarily I integrated this patch, however, as I've been progressing with Swift 3.0 and 0.40.0 I've decided to drop official support for iOS 7.0.

It's a mixture of a need to simplify the code base (iOS 8.0 added activate/deactivate APIs) as well as an effort to reduce support overhead as I'm the only maintainer.

With that in mind, I'm not nefariously trying to make iOS 7.0 support extremely hard, but the final release of 0.40.0 and Swift 3.0 will not be easily compatible.

PR #260, opened 2016-07-25
Joining parts of multi-clause condition

uny:
Small modification to solve some warnings.


robertjpayne:
@uny thanks for this, I had already integrated this just not pushed as I was away. Cheers.

PR #269, opened 2016-08-07
Carthage Support.

uny:
Thank you for your work.

Now, I cannot build this branch with Carthage.
https://github.com/Carthage/Carthage/issues/903


robertjpayne:
@uny does this patch resolve your issues? I'm guessing it does, such a minor tweak though I'm surprised.

uny:
Yes.

> Carthage uses Release configuration to build frameworks, and normally "Enable Testability" is yes only for Debug configuration. I recommend not to build the test target on other than Test action for the library users.

https://github.com/Carthage/Carthage/issues/903#issuecomment-154590239

robertjpayne:
@uny perfect thanks!

PR #272, opened 2016-08-13
support NSView layer-backed view animation

NSLogMeng:
add a new api on updateConstraints for layer-backed view animation

so you can write like this:

``` swift
NSAnimationContext.runAnimationGroup({ (context) in
                    context.duration = 1.0
                    self.view.snp_updateAnimatorConstraints(closure: { (make) in
                        make.width.equalTo(50.0)
                    })
                }, completionHandler: nil)  
```

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -220,7 +220,7 @@ internal class ConcreteConstraint: Constraint {
         self.label = label
     }
     
-    internal func installOnView(updateExisting updateExisting: Bool = false, file: String? = nil, line: UInt? = nil) -> [LayoutConstraint] {
+    internal func installOnView(updateExisting updateExisting: Bool = false, forAnimator: Bool = false, file: String? = nil, line: UInt? = nil) -> [LayoutConstraint] {
         var installOnView: View? = nil
         if self.toItem.view != nil {
             installOnView = closestCommonSuperviewFromView(self.fromItem.view, toView: self.toItem.view)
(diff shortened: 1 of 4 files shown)


NSLogMeng:
Or could add animator on ConstraintDescriptionExtendable?
so we can write like this:

```swift
NSAnimationContext.runAnimationGroup({ (context) in
                    context.duration = 1.0
                    self.view.snp_updateAnimatorConstraints(closure: { (make) in
                        make.width.animator.equalTo(50.0)
                    })
                }, completionHandler: nil)  
```

NSLogMeng:
Or do like this in Masonry:

```
   //store the contraint
    [self.testView mas_makeConstraints:^(MASConstraintMaker *make) {
        make.leading.top.equalTo(self.view);
        self.testConstraint = make.width.equalTo(@100.0);
        make.height.equalTo(@100.0);
    }];

   // update it using animator proxy
    [NSAnimationContext runAnimationGroup:^(NSAnimationContext * _Nonnull context) {
        context.duration = 1.0f;
        self.testConstraint.animator.offset = 50.0f;
    } completionHandler:nil];
```

Anyway, seems there should have a way to get **animator** proxy so that we can use SnapKit on animate.

robertjpayne:
@NSLogMeng on iOS you simply have to run `view.layoutIfNeeded()` inside the animation block, does this not work on macOS at all? I probably need to do some more tests with it if not...

NSLogMeng:
@robertjpayne Hi, I‘m afraid there's some different autolayout based animation between iOS and macOS.
I write a simple [demo](https://github.com/NSLogMeng/AutoLayoutAnimationDemo.git) based on my fork and myself commit, discuss some situation about this problem.
As a reference, Masonry has implement a way to resolve the problem I talked above, but I think it probably not perfect (user mast store the constraint out, when animation , get the constraint.animator and update constant).
thx

robertjpayne:
Closing this as #337 supersedes this

PR #279, opened 2016-08-23
Add snp_alignEdges for quick edge alignment of views

igorefremov:
This allows for faster edge alignment because rather than using
```
view.snp_makeConstraints { (make) in
    make.edges.equalTo(superview)
}
```
users can use
```
view.snp_alignEdges(secondView: superview)
```


robertjpayne:
@igorefremov while I understand the convenience at this time I'd prefer to keep SnapKit's API to just the `snp_<make/remake/update>Constraints`, these shortcuts are best kept inside someones personal code library with a simple SnapKitExtensions.

Issue #282, opened 2016-08-26
Constraint.update(offset:) won't update constant. feature/0.40.0


uny:
### New Issue Checklist

* [x] I have looked at the [Documentation](http://snapkit.io/docs)
* [x] I have read the [F.A.Q.](http://snapkit.io/faq)

### Issue Info

 Info                    | Value                               |
-------------------------|-------------------------------------|
 Platform                | ios
 Platform Version        | 10.0
 SnapKit Version         | feature/0.40.0
 Integration Method      | carthage
 

### Issue Description
I want to update constraint constant.
[This line](https://github.com/uny/SnapKitDebug-282/blob/ac15a2409ff1eb21dc8c98dcc1cd83ee5b0788da/SnapKitDebug/ViewController.swift#L34) doesn't work properly.
I debugged for a while, but I cannot found the solution.
I inserted `updateConstraintsIfNeeded` and `layoutIfNeeded`, but nothing changed.

This is an issue for beta. Can you help me?

robertjpayne:
@uny can you please post your constraint code?

uny:
I uploaded a sample repository.
https://github.com/uny/SnapKitDebug-282/blob/master/SnapKitDebug/ViewController.swift#L34
Is this enough as the answer?

uny:
Debug:

* Using LayoutConstraint directly ([`usingSnapKit `](https://github.com/uny/SnapKitDebug-282/blob/c95cce14bc55ae984ce7180324f561dfa3398161/SnapKitDebug/ViewController.swift#L13) is `false`), it works properly.
* `Constraint.layoutConstraints` become empty.
  1. Make [`Constraint.layoutConstraints`])(https://github.com/uny/SnapKitDebug-282/blob/c95cce14bc55ae984ce7180324f561dfa3398161/Carthage/Checkouts/SnapKit/Source/Constraint.swift#L49) public
  2. [After `makeConstraints`](https://github.com/uny/SnapKitDebug-282/blob/c95cce14bc55ae984ce7180324f561dfa3398161/SnapKitDebug/ViewController.swift#L29), it has the constraint.
  3. [In `didTapButton`](https://github.com/uny/SnapKitDebug-282/blob/c95cce14bc55ae984ce7180324f561dfa3398161/SnapKitDebug/ViewController.swift#L46), however, it becomes empty.

uny:
[...]

PR #285, opened 2016-08-29
Make ConstraintDescription.constraint lazy

uny:
https://github.com/SnapKit/SnapKit/issues/282

This is the fix for the issue.
The problem is solved, but I don't know it is right fix for whole consistency.
Please check it.


robertjpayne:
@uny good catch! I'll need to double check that there isn't a possibility the DSL can continue to be chained after getting a constraint reference though.

uny:
@robertjpayne Thank you!

PR #287, opened 2016-08-30
Added Support for UILayoutGuide

PSchu:
I added Support for UILayoutGuide, which was added with iOS9
It is not the same as topLayoutGuide/bottomLayoutGuide of ViewControllers.
UILayoutGuide objects can be used as dummy Views for Autolayout.
I needed it for the readableContentGuide of UIView.

The only thing wich was strange for me is that the Test produced 12 constraints.
(the similar test for topLayoutGuide of the VC produced only 6, but i normally would expect only two)


robertjpayne:
@PSchu this is already tracking in the feature/0.40.0 branch but understand that that only supports Swift 3.0.

robertjpayne:
Actually @PSchu this is much more complete. Thanks for the additions!

PSchu:
Thanks, needed that still in Swift 2.2. 
We probably can't switch to Swift 3.0 directly

robertjpayne:
@PSchu 👍 I just merged in some good support into Swift 3 branch too.

PR #288, opened 2016-08-31
Fixed testSizeConstraints test as per suggestion

madhavajay:
https://github.com/SnapKit/SnapKit/pull/243


robertjpayne:
@madhavajay thank you! 🎉

PR #323, opened 2016-09-20
fix "Mutating a priority from required to not on an installed constra…

vhosune:
on iOS 8, SnapKit crash with 

*** Assertion failure in -[SnapKit.LayoutConstraint setPriority:], /SourceCache/Foundation_Sim/Foundation-1144.17/Layout.subproj/NSLayoutConstraint.m:174

*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Mutating a priority from required to not on an installed constraint (or vice-versa) is not supported.  You passed priority 1000 and the existing priority was 1000.'

when calling updateOffset(amount:)

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -232,7 +232,7 @@ public class Constraint {
             #if os(iOS)
                 let requiredPriority: UILayoutPriority = UILayoutPriorityRequired
             #else
-                let requiredPriority: UILayoutPriority = 1000.0
+                let requiredPriority: Float = 1000.0
             #endif
             
             
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -229,11 +229,9 @@ public class Constraint {
             let attribute = (layoutConstraint.secondAttribute == .notAnAttribute) ? layoutConstraint.firstAttribute : layoutConstraint.secondAttribute
             layoutConstraint.constant = self.constant.constraintConstantTargetValueFor(layoutAttribute: attribute)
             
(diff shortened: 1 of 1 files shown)


Ekhoo:
👍

Ekhoo:
I have the same crash. When do you think you merge this PR and release a new version ?

robertjpayne:
@vhosune this is only half complete, if we're going to allow updating the priority there must be a check that ensures you are not moving from required to non required as per Apple's documentation:

```
Priorities may not change from nonrequired to required, or from required to nonrequired. An exception will be thrown if a priority of NSLayoutPriorityRequired in macOS or UILayoutPriorityRequired in iOS is changed to a lower priority, or if a lower priority is changed to a required priority after the constraints is added to a view. Changing from one optional priority to another optional priority is allowed even after the constraint is installed on a view.
```

vhosune:
updated with checks

Source/Constraint.swift
```diff
@@ -229,7 +229,14 @@ public class Constraint {
             let attribute = (layoutConstraint.secondAttribute == .notAnAttribute) ? layoutConstraint.firstAttribute : layoutConstraint.secondAttribute
             layoutConstraint.constant = self.constant.constraintConstantTargetValueFor(layoutAttribute: attribute)
             
-            if (layoutConstraint.priority < UILayoutPriorityRequired), (self.priority.constraintPriorityTargetValue != UILayoutPriorityRequired) {
+            #if os(iOS)
```
robertjpayne: Just a small tweak to `#if os(iOS) || os(tvOS)` please for tvOS users.

PR #336, opened 2016-09-24
change the result constraint of maker.center.equalTo(CGPoint) to based on the original point of superview

LiVincent-Zhang:


robertjpayne:
@LiVincent-Zhang I'll need to look at this further but from first glance I don't believe this will allow for the proper behavior when relating to another view explicitly e.g. `make.center.equalTo(otherView)`

If the `from` attribute is `centerX` it should *only* go to `left` if the `equalTo` is implied superview.

LiVincent-Zhang:
@robertjpayne Yes, I realized this is not correct.

LiVincent-Zhang:
@robertjpayne  can you review the new commit?

robertjpayne:
@LiVincent-Zhang thanks!

Issue #294, opened 2016-09-09
equalToSuperview missing


tcurdt:
### New Issue Checklist

* [x] I have looked at the [Documentation](http://snapkit.io/docs)
* [x] I have read the [F.A.Q.](http://snapkit.io/faq)

### Issue Info

 Info                    | Value                               |
-------------------------|-------------------------------------|
 Platform                | ios
 Platform Version        | 10.0
 SnapKit Version         | 0.40 (2e4cebf41d8bd0833487381590ed90c38b9a2491)
 Integration Method      | carthage
 

### Issue Description

```
       myView.snp.updateConstraints { make in
            make.edges.equalToSuperview()
        }
```

Does not compile. It seems the method `equalToSuperview` is missing.

skyebook:
I just ran in to the same thing that you did and wasn't able to find the code being deleted within the `0.4.0` branch. Turns out that this is based off of `0.19.0` (`equalToSuperview` was added for `0.21.0`).

Luckily I don't have too many uses of it since that function was only introduced a few months ago, but this would definitely be a pain if my entire project were built with it.

robertjpayne:
I've purposely left this API out of SnapKit 0.40.0 for now, thinking back it's probably useful to have though in reality all it is doing is doing `myView.superview!` but I suppose that is a nice to have shortcut for many users.

tcurdt:
IMO it's a very nice shortcut ... but I guess one could also add this via extension.

robertjpayne:
Fixed on branch, not tagged yet though.

txaiwieser:
I just updated my project to swift 3.0 and I'm really missing this shortcuts too.
```
lessThanOrEqualToSuperview()
greaterThanOrEqualToSuperview()
```
I'm using as extension but it would be great if added to the lib.

Thanks

PR #341, opened 2016-09-29
Added convenience methods to ConstraintMakerRelatable for superview access

txaiwieser:
Added two new convenience methods to ConstraintMakerRelatable for superview access
`greaterThanOrEqualToSuperview`
`lessThanOrEqualToSuperview`

As mentioned in the issue #294


robertjpayne:
@txaiwieser thanks!

PR #342, opened 2016-10-03
Add `topMargin` to constraint maker

aethe:


robertjpayne:
@aethe could you please make this PR against `develop` instead of `feature/0.40.0`?

Issue #324, opened 2016-09-20
Feature request: Priority enum


Shehryar:
### New Issue Checklist
- [x] I have looked at the [Documentation](http://snapkit.io/docs)
- [x] I have read the [F.A.Q.](http://snapkit.io/faq)
- [x] I have filled out this issue template.
### Issue Info

| Info | Value |
| --- | --- |
| Platform | ios/osx/tvos |
| Platform Version | 9.0 |
| SnapKit Version | 3.0.0 |
| Integration Method | carthage/cocoapods/manually (all) |
### Issue Description

It's not too obvious on how priorities should be used and what values corresponded to the old `priorityLow` - `priorityRequired` without searching through the implementation.

In my opinion creating an enum as simple as:

```
 public enum Priority: Float {
        case low              = 200.0
        case medium      = 600.0
        case high            = 800.0
        case required     = 1000.0
    }
```

would remove a lot of ambiguity for what constitutes relative priority and what makes something `required`. I do think having the flexibility of being able to put on any priority yourself is very helpful, but not knowing the threshold for a constraint being required is unhelpful, so having these enums as just a wrapper for specific values could alleviate that problem.

Additionally adding more documentation around priorities would be helpful as well as there is currently only one example

robertjpayne:
@Shehryar interesting, I specifically removed the priorityLow/High etc… because they mean very different things on macOS and iOS.

An enum may solve that though because it could be specific to the platform.

Shehryar:
I would be happy to open a PR for it myself. If you have a specific way you would like it to be implemented I'd love to help out and contribute

MP0w:
I like the idea of using an enum but IMHO it's often needed to use other priorities:
e.g `low + 1`, `700`.

With this approach we would need to have 2 sets of APIs one that takes `Priority` and the other that takes a `float` and to create priorities like `low + 1` you would probably need: `.low.rawValue + 1`... not a fan of this...

So I would enhance the proposed enum in this way:

```swift
public enum Priority: Strideable, RawRepresentable, Equatable {
    case low
    case medium
    case high
    case required
    case custom(value: Float)
    
    public init?(rawValue: Float) {
        // maybe needs some constarints like 0 <= x <= 1000 ?
        self = .custom(value: rawValue)
    }
    
    public var rawValue: Float {
       // eventually different in iOS & macOS
        switch self {
        case .low:
            return 250
        case .medium:
            return 500
        case .high:
            return 750
        case .required:
[...]

PR #345, opened 2016-10-04
Priority enum

Shehryar:
This adds an enum for priority based constraints allowing users to write priority as:

```
.priority(level: .low)
```

Based on the issue https://github.com/SnapKit/SnapKit/issues/324


Source/ConstraintMakerPriortizable.swift
```diff
@@ -30,38 +30,49 @@
...
+    public enum Priority: Strideable, RawRepresentable, Equatable {
+        case low
+        case medium
+        case high
+        case required
+        case custom(value: Float)
+        
+        public init?(rawValue: Float) {
+            // TODO:
+            // constrain to 0 <= x <= 1000
```
MP0w: since it's a failable initializer we should return nil for invalid priority (not completely sure that <0 or >1000 is invalid...)

Source/ConstraintMakerPriortizable.swift
```diff
@@ -30,38 +30,49 @@
...
+            self = .custom(value: rawValue)
+        }
+        
+        public var rawValue: Float {
+            switch self {
+            case .low:
+                return 250.0
+            case .medium:
+                return 500.0
```
MP0w: looks like was 501 for macOS?

Source/ConstraintMakerPriortizable.swift
```diff
@@ -30,38 +30,49 @@
...
     @discardableResult
-    public func priorityHigh() -> ConstraintMakerFinalizable {
-        return self.priority(750)
+    public func priority(level: Priority) -> ConstraintMakerFinalizable {
+        return self.priority(level: ConstraintMakerPriortizable.Priority(rawValue: level.rawValue)!)
     }
-    
-    @available(*, deprecated:3.0, message:"Use priority(_ amount: ConstraintPriorityTarget) instead.")
-    @discardableResult
-    public func priorityMedium() -> ConstraintMakerFinalizable {
```
MP0w: Maybe we have to keep (but change the implementation) those deprecated methods? Otherwise will need to be a Breaking Change version

robertjpayne:
I agree with most of @MP0w's comments

Shehryar:
Will update this PR soon

Shehryar:
PR is updated! @robertjpayne @MP0w

Shehryar:
still the same `exit code 65` error

Shehryar:
TODO: Need to update this again so the old methods can use the new enum values instead of hardcoded values.

Source/ConstraintMakerPriortizable.swift
```diff
@@ -30,12 +30,68 @@
...
+        
+        public init?(rawValue: Float) {
+            if rawValue < 0 || rawValue > 1000 {
+                return nil
+            }
+            
+            self = .custom(value: rawValue)
+        }
+        
+        var macOSMedium: Float {
```
MP0w: I would make it private or use it inline directly. The naming could be something different ? or is this because other platform don't really have medium priority? (AFAIK)

Shehryar: It's mainly because the other platform don't really have a medium priority

Shehryar:
@robertjpayne please re-review!

MP0w:
Any update?

Source/ConstraintMakerPriortizable.swift
```diff
@@ -30,38 +30,90 @@
 
 public class ConstraintMakerPriortizable: ConstraintMakerFinalizable {
     
+    public enum Priority: Strideable, RawRepresentable, Equatable {
```
MP0w: Oh, IMHO this enum should not be inside ConstraintMakerPriortizable.
We don't want to write ConstraintMakerPriortizable.Priority right?

Source/ConstraintMakerPriortizable.swift
```diff
@@ -30,38 +30,90 @@
...
     @discardableResult
+    @available(*, deprecated:3.0, message:"Use priority(level: .custom(Float)) instead.")
     public func priority(_ amount: ConstraintPriorityTarget) -> ConstraintMakerFinalizable {
         self.description.priority = amount
         return self
     }
     
-    @available(*, deprecated:3.0, message:"Use priority(_ amount: ConstraintPriorityTarget) instead.")
+    @discardableResult
+    public func priority(level: Priority) -> ConstraintMakerFinalizable {
```
MP0w: Not a fan of the label level.
.priority(.low) feels the natural candidate

Source/ConstraintMakerPriortizable.swift
```diff
@@ -30,38 +30,90 @@
...
+        return self.priority(level: ConstraintMakerPriortizable.Priority(rawValue: level.rawValue)!)
```
MP0w: Isn't this a loop??! (It's late but really looks like)
Also don't see the point of initializing a priority with another priority...

robertjpayne:
I'll get around to this sorry! I've been insanely busy.

MP0w:
@robertjpayne Hey! Sorry to push but I feel like this improves the APIs and I'm missing it.
Let me know if I can help with anything to merge this pull request.
I also have a couple of other things that would like to be improved or make it easier to use, will open issues ASAP

robertjpayne:
Hey everyone @MP0w included I've merged this but I've also made some tweaks:

#### Moved the Priority enum to a `ConstraintPriority` struct

There's good reasons to make this a struct rather than an enum. For starters much like enums static members on a struct can be used shorthand:

```
public struct ConstraintPriority {
    public static var required: ConstraintPriority {...}
}
...
public func priority(_ amount: ConstraintPriority)
...
.priority(.required)
```

Using a struct also allows users to define their own cases e.g.:

```
extension ConstraintPriority {
    public var almostRequired: ConstraintPriority { return 999.0 }
}
```

And ConstraintPriority conforms to expressible by float literal so you do not have to use an initialiser to create it from a float literal value.

#### Left the priority function that accepts a ConstraintPriorityTarget

I've left this API in for niceties to not break people's code. This is still useful because anything can conform to that protocol and be allowed in the priority function.


Thanks again for your contribution! 👍

robertjpayne:
Also I removed the stride able conformance, I can't see how that is actually useful? Is there a solid use case for that?

MP0w:
I see the change to struct, the reason for strideable is that you are able to use for example priority(.low + 1), this improves a lot the usage IMHO.
I tested it back then and it worked

MP0w:
That was the implementation https://github.com/SnapKit/SnapKit/issues/324#issuecomment-249150601
Might need adaption with a struct

robertjpayne:
@MP0w I couldn't get it to work w/o ambiguous API errors on the struct, I think it worked on the enum because enum cases are slightly different than static members for shorthand mode.

MP0w:
I will try it out, otherwise would you reconsider using an enum? It can still be extended with static vars that return .custom(value) if the only argument was the extensibility

MP0w:
@robertjpayne I tested it in a playground and seems to work https://gist.github.com/MP0w/4cd6df5558f8db2729035c0a25203398
Do you remember any specific issue? Should I try to create a PR?

PR #357, opened 2016-10-11
add @available to ConstraintLayoutGuide extension

LiVincent-Zhang:
It seems that without @available(iOSApplicationExtension 9.0, *) above the ConstraintLayoutGuide extension the instance of ConstraintLayoutGuide have no member named "makeConstraints".

I use SnapKit in latest release XCode8 and swift3

### DIFF
--- a/Source/LayoutConstraintItem.swift
+++ b/Source/LayoutConstraintItem.swift
@@ -31,6 +31,7 @@
 public protocol LayoutConstraintItem: class {
 }
 
+@available(iOSApplicationExtension 9.0, *)
 extension ConstraintLayoutGuide : LayoutConstraintItem {
 }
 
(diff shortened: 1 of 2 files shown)


robertjpayne:
@LiVincent-Zhang thanks for this PR does it not work with `@available(iOS 9.0, *)`? I don't think it should use the `iOSApplicationExtension` part?

robertjpayne:
@LiVincent-Zhang gah sorry it also needs macOS support so should be:

`@available(iOS 9.0, OSX 10.11, *)` and also a bit lower in the file there is a `#available(iOS 9.0, *)` that needs to be updated to `#available(iOS 9.0, OSX 10.11, *)`

LiVincent-Zhang:
OKay. the error of CI is confused me. Orz..

PR #366, opened 2016-11-09
Fix tests on macOS

eddiekaiger:
The test for `UILayoutGuide` should only run on iOS and tvOS, so I added a preprocessor directive. Also did some cleanup to the test function.


Shehryar:
+1 Thanks! Was just about to look into fixing the tests so my PR could pass

robertjpayne:
Thanks 👍

robertjpayne:
I probably prematurely merged this -- the tests are valid for macOS too they just need to use NSViewController and NSView instead! but I'll fix that later.

PR #368, opened 2016-11-14
Add activated and deactivated properties to Constraint class.

zetasq:
I have a situation where I need to find whether a Constraint object's constraints are all activated or deactivated, so these properties added are really convenient for me :)


robertjpayne:
@zetasq mind sharing your use case a bit more? Just trying to understand before I make this API addition, seems like it may just be an edge case that is easily handled as an extension at the app level when needed?

zetasq:
@robertjpayne My use case is like this:

    private var countryCodeButtonWidthConstraint: Constraint?

    private var isCountryCodeButtonHidden: Bool {
        get {
            return countryCodeButton.isHidden
        }
        set {
            countryCodeButton.isHidden = newValue
            newValue ? countryCodeButtonWidthConstraint?.activate() : countryCodeButtonWidthConstraint?.deactivate()
        }
    }

Without this extension, I have to add **isCountryCodeButtonHidden** to track whether **countryCodeButtonWidthConstraint** is activated or not ( I use **countryCodeButtonWidthConstraint** to control whether the button shows up). If this extension is added, I would need only **countryCodeButtonWidthConstraint** to control the behavior.

robertjpayne:
@zetasq thanks for your PR, I'm adding a different single API:

```
public var isActive: Bool
```

Which will return true if one or more of the layout constraints backed by the constraint is active. We don't need a deactivated as you can use the opposite of isActive.

PR #372, opened 2016-11-21
Make ConstraintMakerRelatable targets work with UILayoutGuide

lapfelix:
This lets us use the `make` DSL with UILayoutGuides with "inferred" constraint descriptions. Example:
```stackView.snp.makeConstraints { (make) in
    make.left.equalTo(mainView.readableContentGuide.snp.left)
    make.right.equalTo(mainView.readableContentGuide.snp.right)
}
```

To this:
```stackView.snp.makeConstraints { (make) in
    make.left.right.equalTo(mainView.readableContentGuide)
}
```

I could probably add more tests, if needed.
The main problem I can see is that [UILayoutGuide](https://developer.apple.com/reference/uikit/uilayoutguide) supports less NSLayoutAnchors than [UIView](https://developer.apple.com/reference/uikit/uiview), specifically, it does not support `firstBaselineAnchor` and `lastBaselineAnchor`. Maybe I could modify `ConstraintMakerRelatable` to make some properties only work with `ConstraintView`, not `ConstraintLayoutGuide` ?


robertjpayne:
@lapfelix I'll have a look at this thanks 👍

Source/ConstraintMakerRelatable.swift
```diff
@@ -57,6 +57,9 @@ public class ConstraintMakerRelatable {
         } else if let other = other as? ConstraintConstantTarget {
             related = ConstraintItem(target: nil, attributes: ConstraintAttributes.none)
             constant = other
+        } else if #available(iOSApplicationExtension 9.0, OSXApplicationExtension 10.11, *), let other = other as? ConstraintLayoutGuide {
```
robertjpayne: Should be just `#available(iOS 9.0, OSX 10.11, *)`

Source/ConstraintRelatableTarget.swift
```diff
@@ -60,3 +60,8 @@ extension ConstraintItem: ConstraintRelatableTarget {
 
 extension ConstraintView: ConstraintRelatableTarget {
 }
+
+@available(OSXApplicationExtension 10.11, *)
```
robertjpayne: Again should just be `@available(iOS 9.0, OSX 10.11, *)`

lapfelix:
Fixed !

Also regarding the availability of the `firstBaselineAnchor` and `lastBaselineAnchor` NSLayoutAnchors on UILayoutGuide, should that be fixed ? (Accessing them wouldn't cause a crash: it simply wouldn't work)

robertjpayne:
@lapfelix sorry incredibly busy these days, I'll have a look at the baseline stuff

PR #382, opened 2016-12-19
Restores access to native constraints via 'layoutConstraints' property

colinta:
I tried to follow the git log of how/why this was changed from `public` to `private`, but I wasn't able to connect the history of `Source/Constraint.swift` that far back.  I'm hoping this was an accidental refactor!

Thanks

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -28,10 +28,10 @@
 #endif
 
 public final class Constraint {
-    
+
     internal let sourceLocation: (String, UInt)
     internal let label: String?
-    
+
     private let from: ConstraintItem
     private let to: ConstraintItem
     private let relation: ConstraintRelation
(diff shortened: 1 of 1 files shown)


robertjpayne:
@colinta do you have a use case you can give me for why this should be public again? I believe I made it private once I added proper update methods as capturing the native constraints can mess with SnapKit's internals.

colinta:
We need to pass the `NSLayoutConstraint` to a system that doesn't use/support SnapKit.  We build the constraints with SnapKit, then grab this one constraint out (using `layoutConstraints.first`).

Since this property was public in a previous release (and I introduced the PR that added this property in the first place) I was surprised that it was changed to private.

n-studio:
Use case:

```
var constraint: Constraint?
view.snp.makeConstraints({ (make) in
  constraint = make.top.equalTo(10).constraint
})

// move view by 10 pixels from previous position
view.snp.updateConstraints({ (make) in
  make.top.equalTo(constraint!.constant + 10)
  // or
  make.top.equalTo(constraint!.layoutConstraints[0].constant + 10)
})
```

robertjpayne:
@colinta fair enough! Sorry I probably forgot and was curious as to why it was public. I'll let this go public for now and hopefully few people abuse it too much.

PR #387, opened 2017-01-05
Fix wrong function call within ConstraintViewDSL

YusakuNo1:
The setter of ConstraintViewDSL::contentCompressionResistanceHorizontalPriority calls setContentHuggingPriority instead of setCompressionResistancePriority, I think it's a mistake


robertjpayne:
@YusakuNo1 thank you! It indeed does call the wrong method

PR #391, opened 2017-01-12
Switched main target build settings "skip install"  to YES.

tehjord:
I believe setting the "skip install" settings to true in your project's target's build settings is the standard practice for libraries. Otherwise, if you try and embed the library's target into your project's main target, you might run into an issue whereas an xcode generic archive is created instead of an IPA file when archiving your project. I had this exact problem just now since I'm using Carthage and manually integrating the Snapkit target into my project's main target.


tehjord:
Are You guys not interested in that fix? It is likely messing up build for a lot of people manually integrating snapkit and using fastlane.

robertjpayne:
@tehjord apologies not necessarily not interested just very very busy at the moment.

PR #400, opened 2017-02-01
Set SKIP_INSTALL = YES for SnapKit target

kouky:
Resolves an xcodebuild issue when SnapKit is a subproject, xcodebuild fails to construct a valid app archive for the main project. See [reference SO post](http://stackoverflow.com/questions/32841300/xcodebuild-exportarchive-exportoptionsplist-error-for-key-method-expected-o
)

Also refer to the [template framework xconfig](https://github.com/jspahrsummers/xcconfigs/blame/3d9d99634cae6d586e272543d527681283b33eb0/Base/Targets/Framework.xcconfig) file from [jspahrsummers/xcconfigs](https://github.com/jspahrsummers/xcconfigs/) - the git blame for the `SKIP_INSTALL = YES` change states

> Enables the framework to be included from any location as long as the loader’s runpath search paths includes it. For example from an application bundle (inside the "Frameworks" folder) or shared folder


nickdomenicali:
There is already a PR for this. #391

kouky:
Ok closing this PR, it'd be great to get #391 merged

PR #432, opened 2017-07-07
Add methods to position views relative to each other in a more easy way.

LcTwisk:
Hi! I'm often creating layouts where views are placed in a certain order (e.g top to bottom) and in order to place one view under the order we now have to set constraints like this:

```swift
viewB.snp.makeConstraints { (make) in
    make.top.equalTo(viewA.snp.bottom)
}
```

Which is quite simple, but I think it would be more readable if you could do it like this:

```swift
viewB.snp.makeConstraints { (make) in
    make.under(viewA)
}
```

So I've added a few methods: 'under, above, left:of and right:of' to support this. What do you think?

### DIFF
--- a/Source/ConstraintMaker.swift
+++ b/Source/ConstraintMaker.swift
@@ -143,6 +143,26 @@ public class ConstraintMaker {
         return self.makeExtendableWithAttributes(.centerWithinMargins)
     }
     
+    @discardableResult
+    public func under(_ other: ConstraintView, _ file: String = #file, _ line: UInt = #line) -> ConstraintMakerEditable {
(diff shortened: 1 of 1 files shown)


bleeckerj:
I too have views with a consistent layout like this, typically aligned on a grid. This could be useful. My 2¢.

robertjpayne:
I really appreciate the PR but one of the core principles of SnapKit is to be a little more verbose in order to ensure clarity.

Layout code is very hard to read already and adding convenience rather than sticking with a single way of doing this I believe makes it harder to read and a little more ambiguous.

I'm closing this due to this, but I do appreciate your PR.

PR #445, opened 2017-09-05
Swift 4 Update

joeblau:
Looks like there are more discrepancies between UIKit and AppKit.  I modified all of the places in the codebase where `NSLayout` and `UILayout` differ. There are also general formatting changes.  I also updated the unit tests.


robertjpayne:
I don't think this is necessary after merging #446 which I believe simplified this a lot and is a cleaner solution but I do appreciate your PR.

joeblau:
Ah yep! Totally agree. I'll reset my remote back to this repo.  Thanks for keeping this maintained.

Issue #448, opened 2017-09-13
Document how to use Safe Area Layout Guide iOS 11


sprint84:
### New Issue Checklist

* [x] I have looked at the [Documentation](http://snapkit.io/docs)
* [x] I have read the [F.A.Q.](http://snapkit.io/faq)
* [x] I have filled out this issue template.

### Issue Info

 Info                    | Value                               |
-------------------------|-------------------------------------|
 Platform                | e.g. ios
 Platform Version        | e.g. 11.0
 SnapKit Version         | e.g. 3.0.0
 Integration Method      | e.g. carthage/cocoapods/manually
 

### Issue Description

iOS 11 introduced a new layout guide, Safe Area Layout Guide, which is meant to replace the top/bottom layout guides. This is required for the new iPhoneX

https://developer.apple.com/ios/human-interface-guidelines/overview/iphone-x/

marclove:
This is possible already, though some documentation could be helpful. `safeAreaLayoutGuide` is a `UILayoutGuide`, which is already relatable in SnapKit via `snp`. Example:

```swift
let subview = UIView()
addSubview(subview)
subview.snp.makeConstraints { make in
  if #available(iOS 11, *) {
    make.top.equalTo(safeAreaLayoutGuide.snp.top)
  } else {
    make.top.equalTo(self)
  }
}
```

robertjpayne:
@sprint84 @marclove is right here, you can already do this out of the box, and I don't think there's really a better way to do this because the safe area layout guides are on a view controller rather than a view and we don't extend view controller APIs

robertjpayne:
This is now in the docs, but I'm going to leave this issue open for awhile because it will prevent a bajillion new ones as people start to support safe layout areas.

phamquochoan:
As of Xcode 9 GM, `safeAreaLayoutGuide` has been moved to `UIView`'s property.

https://developer.apple.com/documentation/uikit/uiview

robertjpayne:
@phamquochoan it's still a UILayoutGuide and an already accessible property. I suppose it could make sense to do something like `.snp.safeArea` but it may be confusing of whether that is hitting the view controller or the view itself…

phamquochoan:
I just check the documentation of `UIView` and `UIViewController`, the `safeAreaLayoutGuide` is removed from `UIViewController`

`UIViewController`
```
var additionalSafeAreaInsets: UIEdgeInsets
Custom insets that you specify to modify the view controller's safe area.
[...]

PR #456, opened 2017-09-20
Add compatibility constraints for safe area.

rwinzhang:

PR #459, opened 2017-09-22
Add safeArea to ConstraintViewDSL

bjarkehs:
Hi,

Wanted to implement what was discussed in #448. I made it default to the view if prior to iOS 11. This being because I figured if you want to target an iPhone X, you don't really want to wrap every call to `safeArea.leading` and `safeArea.trailing` in `if #available(iOS 11, *) {`.

It does cause some confusion for iOS 10 and below since you'll still have to use topLayoutGuide on the viewController. But I couldn't figure a way to do that in a good way. I'm still thinking of a way to add an extension that makes sense.

### DIFF
--- a/SnapKit.xcodeproj/project.pbxproj
+++ b/SnapKit.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		2DBA080E1F1FAD66001CFED4 /* Typealiases.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2DBA080D1F1FAD66001CFED4 /* Typealiases.swift */; };
+		51FD13CC1F76071500B868B8 /* ConstraintSafeAreaDSL.swift in Sources */ = {isa = PBXBuildFile; fileRef = 51FD13CB1F76071500B868B8 /* ConstraintSafeAreaDSL.swift */; };
 		EE235F5F1C5785BC00C08960 /* Debugging.swift in Sources */ = {isa = PBXBuildFile; fileRef = EE235F5E1C5785BC00C08960 /* Debugging.swift */; };
 		EE235F6D1C5785C600C08960 /* Constraint.swift in Sources */ = {isa = PBXBuildFile; fileRef = EE235F621C5785C600C08960 /* Constraint.swift */; };
 		EE235F701C5785C600C08960 /* ConstraintDescription.swift in Sources */ = {isa = PBXBuildFile; fileRef = EE235F631C5785C600C08960 /* ConstraintDescription.swift */; };
@@ -48,6 +49,7 @@
 
 /* Begin PBXFileReference section */
 		2DBA080D1F1FAD66001CFED4 /* Typealiases.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Typealiases.swift; sourceTree = "<group>"; };
+		51FD13CB1F76071500B868B8 /* ConstraintSafeAreaDSL.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ConstraintSafeAreaDSL.swift; sourceTree = "<group>"; };
 		537DCE9A1C35CD4100B5B899 /* UIKit.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = UIKit.framework; path = Platforms/AppleTVOS.platform/Developer/SDKs/AppleTVOS9.1.sdk/System/Library/Frameworks/UIKit.framework; sourceTree = DEVELOPER_DIR; };
 		EE235F5E1C5785BC00C08960 /* Debugging.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = Debugging.swift; sourceTree = "<group>"; };
 		EE235F621C5785C600C08960 /* Constraint.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = Constraint.swift; sourceTree = "<group>"; };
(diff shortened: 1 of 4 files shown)


bjarkehs:
One thing I thought about when trying this out in my own project was that it would've been convenient with a `equalToSuperviewSafeArea()`. A bit like #456. But more for convenience. I feel this achieves basically the same as #456, but a very common use case is:
```swift 
make.leading.trailing.equalToSuperview().inset(someInset)
``` 
or similar. 

Compare this:
```swift
make.leading.trailing.equalToSuperviewSafeArea()
```
to this:
```swift
make.leading.equalTo(view.snp.safeArea.leading)
make.trailing.equalTo(view.snp.safeArea.trailing)
```

bjarkehs:
I was playing a bit more with the safeAreas for `UIViewController` and came up with this:

```swift
extension UIViewController {
    var safeAreaTop: ConstraintItem {
        guard #available(iOS 11, *) else {
            return self.topLayoutGuide.snp.bottom
        }
        return self.view.snp.safeArea.top
    }
    
    var safeAreaBottom: ConstraintItem {
        guard #available(iOS 11, *) else {
            return self.bottomLayoutGuide.snp.top
        }
        return self.view.snp.safeArea.bottom
    }
}
```

bjarkehs:
The following code bothered me enough to make me revisit my implementation:

```swift
make.leading.equalTo(view.snp.safeArea.leading)
make.trailing.equalTo(view.snp.safeArea.trailing)
```

With the latest commit this is possible to do as:
```swift
make.leading.trailing.equalTo(view.snp.safeArea)
```

I'm still not quite sure whether the implementation is the right way to do it, but at least now it's there and can be discussed.

robertjpayne:
@bjarkehs I think we may have complicated this a lot here in v4.0.0 you have to use this:

```
make.leading.trailing.equalTo(view.safeAreaLayoutGuide)
```

Given that the entire SnapKit engine already has robust support for Layout Guide's I don't think we need a whole new DSL chain, all we need to do is add a shortcut on `ConstraintViewDSL` that provides quick access to a `ConstraintLayoutGuideDSL` targeting the `safeAreaLayoutGuide`

I could be wrong but at quick glance we should only have to add this to `ConstraintViewDSL` to implement the much shortcut API:

```
    @available(iOS 11.0, *)
    public var safeArea: ConstraintLayoutGuideDSL {
        return ConstraintLayoutGuideDSL(guide: self.view.safeAreaLayoutGuide)
    }
```

bjarkehs:
The reason that I don’t find that solution particularly interesting is that then I have to wrap everything in `if #available(iOS 11.0, *)` like I wrote in my initial post. Meaning adoption will be much more painful. Storyboards and nib files get a new option to simply toggle safe area usage, and I wanted something similar for the programmatic approach.

robertjpayne:
If you do this in storyboards are the constraints then no-op on iOS 10 or do they fallback to something?

bjarkehs:
I'm not entirely sure, since I haven't used storyboards in quite a while. But according to this: https://useyourloaf.com/blog/safe-area-layout-guide/ you can simply activate safe area layouts in interface builder and it'll make it backwards compatible. We can't quite do this from code.

robertjpayne:
@bjarkehs Hmm I see what you're saying here, a lot of users will just want to use "Safe Area" in either situation.

I need to dive into it more.

robertjpayne:
@bjarkehs I still feel like it's smart to force users to use availability checks. If you read the documentation about the safe area layout guide:

> When the view is visible onscreen, this guide reflects the portion of the view that is not covered by navigation bars, tab bars, toolbars, and other ancestor views. (In tvOS, the safe area reflects the area not covered the screen's bezel.) If the view is not currently installed in a view hierarchy, or is not yet visible onscreen, the layout guide edges are equal to the edges of the view.

For the view controller's root view, the layout guide accommodates the status bar, other visible bars, and any additional insets that you specified using the additionalSafeAreaInsets property of your view controller. For other views in the view hierarchy, the layout guide reflects only the portion of the view that is covered by other content. For example, if a view is entirely within the safe area of its superview, the layout guide edges are equal to the edges of the view.

It's actually pretty invalid to assume it's the view bounds on iOS 10 and lower. Because of this I don't want to implement it as a fallback to `view` because it would be ambiguous and confusing.

bhstrifork:
I understand your concern. I'll probably end up using the extension described earlier:

```swift
@available(iOS, introduced: 8.0, deprecated: 11.0, message: "Please use view.snp.safeArea instead")
extension UIViewController {
    var safeAreaTop: ConstraintItem {
        guard #available(iOS 11, *) else {
            return self.topLayoutGuide.snp.bottom
        }
        return self.view.snp.safeArea.top
    }
    
    var safeAreaBottom: ConstraintItem {
        guard #available(iOS 11, *) else {
            return self.bottomLayoutGuide.snp.top
        }
        return self.view.snp.safeArea.bottom
    }
}
```

Maybe adding a `safeAreaLeading` and `safeAreaTrailing`. This would be using the simple DSL you provide, but give me some convenience on projects that target lower than iOS 11.

bjarkehs:
I understand your concern. I'll probably end up using the extension described earlier:

```swift
@available(iOS, introduced: 8.0, deprecated: 11.0, message: "Please use view.snp.safeArea instead")
extension UIViewController {
    var safeAreaTop: ConstraintItem {
        guard #available(iOS 11, *) else {
            return self.topLayoutGuide.snp.bottom
        }
        return self.view.snp.safeArea.top
    }
    
    var safeAreaBottom: ConstraintItem {
        guard #available(iOS 11, *) else {
            return self.bottomLayoutGuide.snp.top
        }
        return self.view.snp.safeArea.bottom
    }
}
```

Maybe adding a `safeAreaLeading` and `safeAreaTrailing`. This would be using the simple DSL you provide, but give me some convenience on projects that target lower than iOS 11.

robertjpayne:
@bjarkehs for now I'm going to leave this as up to the implementor to create convenience accessors for. I really do appreciate your inputs here but unfortunately I don't feel comfortable pushing an assumption that the safe area is always identical to the view's bounds as per Apples documentation.

bjarkehs:
That is perfectly fine. Does it make sense to add the `safeArea` property still? So you can write things like: `view.edges.equalTo(superview.safeArea)`?

PR #473, opened 2017-11-07
[Docs] Update safeAreaLayoutGuide usage - now a property of UIView

beltex:
First, thanks for the great project! :)

Just a minor update to the docs given that `safeAreaLayoutGuide` is no longer a property of `UIViewController`, but instead `UIView`.

**References**

* https://github.com/SnapKit/SnapKit/issues/448#issuecomment-330114911
* [UIView - safeAreaLayoutGuide](https://developer.apple.com/documentation/uikit/uiview/2891102-safearealayoutguide)

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -331,7 +331,7 @@ class MyViewController: UIVewController {
 ```
 
 
-### Snap view to safe layout guide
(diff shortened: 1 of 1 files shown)


robertjpayne:
Thanks!

PR #485, opened 2018-01-05
Add labeled method to docs

tyleralves:
The labeled feature is amazing for debugging, but is not obviously documented.


robertjpayne:
Thanks!

PR #496, opened 2018-02-23
Adds 'constraint.isActive = bool' to Constraint

colinta:
I've got lots of places where I activate one constraint and deactivate another based on some bool, it has been much nicer to use a function that accepts a bool:

```swift
let buttonVisible = someCondition
wideConstraint.isActive = buttonVisible
narrowConstraint.isActive = !buttonVisible
```

Or as a diff to show what I mean:

```diff
 let buttonVisible = someCondition
-if buttonVisible {
-    wideConstraint.activate()
-    narrowConstraint.deactivate()
-else {
-    wideConstraint.deactivate()
-    narrowConstraint.activate()
-}
+wideConstraint.isActive = buttonVisible
+narrowConstraint.isActive = !buttonVisible
```

Function name was the first that popped in my head, but I was following the `update(offset: ..)` naming scheme.  Tests included, plus a minor refactor to `prepare/make/update–Constraints` to delegate identical code to `prepareConstraints`.

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -188,6 +188,15 @@ public final class Constraint {
         self.deactivate()
     }
 
+    public func set(isActivated: Bool) {
+        if isActivated {
+            self.activate()
+        }
+        else {
+            self.deactivate()
+        }
+    }
+
     public func activate() {
         self.activateIfNeeded()
     }
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -49,12 +49,23 @@ public final class Constraint {
     public var layoutConstraints: [LayoutConstraint]
     
     public var isActive: Bool {
-        for layoutConstraint in self.layoutConstraints {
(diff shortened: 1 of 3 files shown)


robertjpayne:
@colinta what do you reckon about a `var isActivated: Bool` instead? I feel like it's cleaner syntax and we can then also read the state?

colinta:
Makes sense! I’ll update the PR soon.

colinta:
I see that there's already a property `isActive`, that feels like the right place to put this:

```swift
    public var isActive: Bool {
        set {
            if newValue {
                activate()
            }
            else {
                deactivate()
            }
        }
        
        get {
            for layoutConstraint in self.layoutConstraints {
                if layoutConstraint.isActive {
                    return true
                }
            }
            return false
        }
    }
```

(PR and specs updated)

robertjpayne:
Thanks @colinta looks good

PR #520, opened 2018-06-21
make ConstraintAttributes conform to ExpressibleByIntegerLiteral

guzhenhuaGitHub:
make:
```Swift
internal static var left: ConstraintAttributes { return self.init(1) } 
```
to:
```Swift
internal static var left: ConstraintAttributes { return 1 } 
```
and so on


robertjpayne:
Very nice!

Issue #498, opened 2018-02-26
centerX, centerY offset/inset are the same


sh3244:
### New Issue Checklist

🚫 If this template is not filled out your issue **will** be closed with no comment. 🚫

* [ X] I have looked at the [Documentation](http://snapkit.io/docs)
* [ X] I have read the [F.A.Q.](http://snapkit.io/faq)
* [ X] I have filled out this issue template.

### Issue Info

 Info                    | Value                               |
-------------------------|-------------------------------------|
 Platform                | iOS
 Platform Version        | 10.0
 SnapKit Version         | 4.0.0
 Integration Method      | cocoapods
 

### Issue Description

It appears that centerY.offset(value) does the same thing that centerY.inset(value) does.
In order to circumvent, one must provide a negative value.

AlexanderBollbach:
yes it does seem that way.  `offset` and `inset` are the same.   this is very confusing and unhelpful

Cyberbeni:
Fixed by #569

PR #524, opened 2018-07-11
CenterX/Y behaviour with inset.

Cyberbeni:
https://github.com/SnapKit/SnapKit/issues/498

Had the same issue as described here, changed inset handling to be more logical for centerX/Y


Cyberbeni:
Changes are included in #569

Issue #504, opened 2018-04-03
Is there a plan to support swift static library?


nspangbo:
### New Issue Checklist

🚫 If this template is not filled out your issue **will** be closed with no comment. 🚫

* [x] I have looked at the [Documentation](http://snapkit.io/docs)
* [x] I have read the [F.A.Q.](http://snapkit.io/faq)
* [x] I have filled out this issue template.

### Issue Info

 Info                    | Value                               |
-------------------------|-------------------------------------|
 Platform                | e.g. ios/osx/tvos
 Platform Version        | e.g. 8.0
 SnapKit Version         | e.g. 0.19.0
 Integration Method      | e.g. carthage/cocoapods/manually
 

### Issue Description

Is there a plan to support swift static library?
http://blog.cocoapods.org/CocoaPods-1.4.0/

robertjpayne:
Yep it sure can be done pretty easy just a small change, would love if you want to make a PR!

0xwangbo:
http://blog.cocoapods.org/CocoaPods-1.5.0/

A better way.

PR #525, opened 2018-07-15
marked as static in podspec

Cyberbeni:
Tested by adding this to an existing project's podfile `pod 'SnapKit', :git => 'https://github.com/Cyberbeni/SnapKit.git', :branch => 'static-framework'` and then running `pod install` and building the app.


Cyberbeni:
https://github.com/SnapKit/SnapKit/issues/504

PR #529, opened 2018-08-11
Refactor ACL and ConstraintAttributes

cruisediary:
- Remove internal access contorl level bacause it is default access control level in Swift 4
> Default Access Levels
All entities in your code (with a few specific exceptions, as described later in this chapter) have a default access level of internal if you don’t specify an explicit access level yourself. As a result, in many cases you don’t need to specify an explicit access level in your code.

https://docs.swift.org/swift-book/LanguageGuide/AccessControl.html

- Refactor ConstraintAttributes using enum's Dot syntax
```swift
// as is
return ConstraintItem(target: self.target, attributes: ConstraintAttributes.centerX)
// to be
return ConstraintItem(target: self.target, attributes: .centerX)
```

### DIFF
--- a/Source/Constraint.swift
+++ b/Source/Constraint.swift
@@ -29,8 +29,8 @@
 
 public final class Constraint {
 
-    internal let sourceLocation: (String, UInt)
-    internal let label: String?
+    let sourceLocation: (String, UInt)
+    let label: String?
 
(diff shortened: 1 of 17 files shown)


cruisediary:
@robertjpayne Please, Any update ? :)

PR #541, opened 2018-09-14
Update  to Swift 4.2

rhysforyou:
This PR updates to SnapKit to build with Swift 4.2, which involves a few key changes:

- Update the Swift version in the project's build settings
- Update to recommended project settings as suggested by Xcode 10
- Removes the `LayoutAttribute` and `LayoutRelation` attributes used for cross platform compatibility, as now all platforms use `NSLayoutConstraint.Attribute` and `NSLayoutConstraint.Relation
- Specifies an explicit Swift version in the project's Podspec
- Updates the Swift Package Manager manifest to the latest format
- Update the Travis CI build script to use Xcode 10

I have Travis running against my fork of the repo, you can see build results here: https://travis-ci.org/rhysforyou/SnapKit


spadafiva:
@rhysforyou I would love to see this added!  Thanks for doing this.

freak4pc:
Looking into the `develop` branch (and also just built for myself on my project) - this was already resolved in #543. Seems like this PR is not needed, let me know if you think otherwise?

aapierce0:
Sorry, I'm confused. It looks like issue #543 wasn't resolved?

In any case, I'm trying to use Swift 4.2 in my project, and as of now, adding SnapKit as a CocoaPods dependency causes my app to not compile. Using https://github.com/rhysforyou/SnapKit/tree/swift-4.2 as a remote dependency resolves it.

freak4pc:
Sorry, I meant #516, and I meant the `develop` branch. 
Can you try adding:

```rb
pod 'SnapKit', git: 'https://github.com/SnapKit/SnapKit', branch: 'develop'
```

aapierce0:
@freak4pc Aha, that seems to work. Thanks!

Kalvin126:
Could we set the `spec.swift_version = '4.2'` in the `SnapKit.podspec` as well?

freak4pc:
I don't see why you'd want to do that. It'd break backward compatibility for Swift 4 projects.
Better to leave as is, in which case it would be based on the Project's version.

Kalvin126:
I mean I'd wish the current podspec had `spec.swift_version = '4.0', because my project is using Swift 4.2, but SnapKit and other pods are not yet 4.2 ready and CocoaPods is resolving to use my project's Swift version which is 4.2.

Panajev:
You can run a post pod install hook adding it in your Podfile and in there force maybe even all your pod to go in Swift 4/4.1 mode... you can do more advanced processing but debugging these kind of changes is a bit of trial and error using the Xcodeproj gem Cocoapods relies on.

Kalvin126:
oh, `spec.swift_version = '4.2'` was actually added in this PR

codeguesser:
image
There must be some problem in it

freak4pc:
@fassko I'm not sure what you mean exactly. I never saw a good reason to use `spec.swift_version` in the podspec level as I feel its a bit aggressive. If the project can be built on both Swift 4 and Swift 4.2, the absence of that is better so the decision is made by the project itself.

fassko:
@freak4pc I think you're right, makes total sense. Thanks!

freak4pc:
4.0.1 was released : https://github.com/SnapKit/SnapKit/releases/tag/4.0.1
I would close this PR If I were you, as it seems Xcode 10 backward-compatible support has been addressed.

mattbarker016:
There's still a warning that SnapKit 4.0.1 needs to be converted to Swift 4.2

mattbarker016:
Why was this closed? There's still a warning that SnapKit 4.0.1 needs to be converted to Swift 4.2.

likeSo:
image

likeSo:
You mean NSLayoutConstraint.Relation can not be used on Swift 4.2?😀 But i can see that in Xcode:
image
SnapKit version in my Podfile is 4.0.1, and my project still needs to migrate:
image
After conversion, that two changes made...

freak4pc:
Made a PR to fix the minor warning bothering y'all :) #547

PR #546, opened 2018-09-26
Modify build settings swift language version to Swift 4.2

IDEAL-YANG:
Use it in myself project. Build failed ! So modify it !


freak4pc:
Can you explain why this is needed? The project should build fine in a backward-compatible way on both Swift 4 and Swift 4.2.

robertjpayne:
I'm with @freak4pc this project shouldn't need an Swift 4.2 flag specifically so for now I'm going to close this

PR #553, opened 2018-10-25
Is it a spelling mistake?

guzhenhuaGitHub:
Class **ConstraintMakerPriortizable** should be **ConstraintMakerPrioritizable**?


robertjpayne:
Thanks! Have fixed this but did it directly to include a Typealias and file rename.

PR #574, opened 2019-03-10
[EXC_BAD_ACCESS] Update LayoutConstraint == operator to support iOS 10 and higher changes

sashabelonogov:
Hello SnapKit team,

Recently, we had a very weird `EXC_BAD_ACCESS (code=EXC_I386_GPFLT)` crash, that was happening on the updating of the constraints randomly but still quite often (maybe once every 4 times the screen was opened). 
The code was crashing on the guard statement of the `==` operator of the `LayoutConstraint` class. After further investigation, we found out that the crash is caused by the `firstItem` or `secondItem` fields of the `NSLayoutConstraint` of the `UIKit`. Digging into the documentation of this class, you can notice that those fields are `unowned(unsafe)`. 
The Apple comment says: 
```
/* accessors
     firstItem.firstAttribute {==,<=,>=} secondItem.secondAttribute * multiplier + constant
     Access to these properties is not recommended. Use the `firstAnchor` and `secondAnchor` properties instead.
     */
```
After replacing `firstItem.firstAttribute` and `secondItem.secondAttribute` with `firstAnchor` and `secondAnchor` the crashes disappeared.
You can see the fix of this issue above. Also, the test is prepared to make sure it works correctly on all the platforms.

Please let me know what you think!

### DIFF
--- a/Source/LayoutConstraint.swift
+++ b/Source/LayoutConstraint.swift
@@ -44,14 +44,21 @@ public class LayoutConstraint : NSLayoutConstraint {
 }
 
 internal func ==(lhs: LayoutConstraint, rhs: LayoutConstraint) -> Bool {
-    guard lhs.firstItem === rhs.firstItem &&
-          lhs.secondItem === rhs.secondItem &&
-          lhs.firstAttribute == rhs.firstAttribute &&
-          lhs.secondAttribute == rhs.secondAttribute &&
-          lhs.relation == rhs.relation &&
-          lhs.priority == rhs.priority &&
-          lhs.multiplier == rhs.multiplier else {
-        return false
+    let areLayoutAnchorsEqual: Bool
+    if #available(iOS 10.0, OSXApplicationExtension 10.12, *) {
+        areLayoutAnchorsEqual = lhs.firstAnchor === rhs.firstAnchor &&
+            lhs.secondAnchor === rhs.secondAnchor
+    } else {
+        areLayoutAnchorsEqual = lhs.firstItem === rhs.firstItem &&
+            lhs.secondItem === rhs.secondItem &&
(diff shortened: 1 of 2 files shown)


Source/LayoutConstraint.swift
```diff
@@ -44,14 +44,21 @@ public class LayoutConstraint : NSLayoutConstraint {
...
+        areLayoutAnchorsEqual = lhs.firstItem === rhs.firstItem &&
+            lhs.secondItem === rhs.secondItem &&
+            lhs.firstAttribute == rhs.firstAttribute &&
+            lhs.secondAttribute == rhs.secondAttribute
+    }
+    guard areLayoutAnchorsEqual &&
+        lhs.relation == rhs.relation &&
+        lhs.priority == rhs.priority &&
+        lhs.multiplier == rhs.multiplier else {
+            return false
```
sergeysmagleev: The `guard` statement seems redundant, shouldn't you just return the result of the comparison?

sashabelonogov: yes, you're right. It will make it way simpler

robertjpayne:
Thanks!

bill350:
@robertjpayne This merge request created regression and broken the 5.0.0 release.
Please check https://github.com/SnapKit/SnapKit/commit/bc2b0b93320513eb747b3baf6332c05bacc067b4

robertjpayne:
So it looks like there's really no good way around the crasher. Unfortunately while Apple's documentation state to use NSLayoutAnchor instead there's zero way to compare to NSLayoutAnchor's and determine if they are the same object.

The root cause of this issue is because you're somehow still referencing views that use constraints generated from items that have been deallocated. The fix is to keep strong references to those items until the views are gone.

For now I have no choice but to rollback this PR as it causes more issues than it solves.

PR #570, opened 2019-01-29
Support swift 5.0

kawoou:
* Support swift 5.0.
 * Update podspec swift version.
 * Fix warning.

PR #575, opened 2019-03-18
Update for Swift 5 compiler and Xcode 10.2

alanzeino:
This commit supports the Swift 5 compiler and Xcode 10.2.

Notes:

* Because Swift 3 is no longer supported by Xcode 10.2, we no longer can use `3.0` as an attribute in `@available`. This was used in a number of places to deprecate the old SnapKit API. In this diff, I deleted that API since I didn't really have any other option. This means the version of SnapKit that contains this commit will need the major version bumped. Happy to talk through this.

From the [docs](https://developer.apple.com/documentation/xcode_release_notes/xcode_10_2_beta_4_release_notes/swift_5_release_notes_for_xcode_10_2_beta_4):

> Swift 3 mode has been removed. Supported values for the -swift-version flag are 4, 4.2, and 5. (43101816)

Example warning:

Screen Shot 2019-03-18 at 2 18 02 PM

* The Swift 5 compiler (in all modes; 4, 4.2 and 5.0) warns about redundant access control levels, so I removed them.

Example warning:

Screen Shot 2019-03-18 at 2 18 20 PM

* Enumerations coming from Objective-C are now required to handle unknown future cases, or the reality that any value can be set inside the type as long as it fits. I handled these like so:

```
switch layoutAttribute {
                case .left, .right, .leading, .trailing, .centerX, .leftMargin, .rightMargin, .leadingMargin, .trailingMargin, .centerXWithinMargins:
                    return value.x
                case .top, .bottom, .centerY, .topMargin, .bottomMargin, .centerYWithinMargins, .lastBaseline, .firstBaseline:
                    return value.y
                case .width, .height, .notAnAttribute:
                    return 0.0
                #if swift(>=5.0)
                @unknown default:
                    return 0.0
                #endif
            }
```

In some cases, I didn't know if there was a value to return that made sense, so I did this:

```
switch relation {
    case .equal:                return "=="
    case .greaterThanOrEqual:   return ">="
    case .lessThanOrEqual:      return "<="
    #if swift(>=5.0)
    @unknown default:
        return "unknown" // If you see this, SnapKit doesn't support a recently added case in this enumeration from Objective-C. It needs to be updated to support the new case. Update your SnapKit version.
    #endif
    }
```

Basically telling whomever is reading the code that a new case was probably added that has made the code return `unknown`. Happy to take your guidance on how to resolve this one.

From the docs:

> In Swift 5 mode, switches over enumerations that are declared in Objective-C or that come from system frameworks are required to handle unknown cases—cases that might be added in the future, or that may be defined privately in an Objective-C implementation file. Formally, Objective-C allows storing any value in an enumeration as long as it fits in the underlying type. These unknown cases can be handled by using the new @unknown default case, which still provides warnings if any known cases are omitted from the switch. They can also be handled using a normal default case.
> 
> If you’ve defined your own enumeration in Objective-C and you don’t need clients to handle unknown cases, you can use the NS_CLOSED_ENUM macro instead of NS_ENUM. The Swift compiler recognizes this and doesn’t require switches to have a default case.


tzvist:
Is this PR scheduled for merge?

(new Xcode was released)

anlaital:
Thank you for all the great Xcode 10.2 / Swift 5 PRs, @alanzeino! 👌

freak4pc:
Seems there was already a PR for this opened about two months ago but is up-to-date (#570), might've been nicer to collaborate on it instead of overriding their work :) Just my $0.02. 

Anyhow, @robertjpayne - do you have any plans for a Swift 5.x, Xcode 10.2 release?

robertjpayne:
Merged previous PR so this is unnecessary now, thank you for the request though I appreciate it.

PR #581, opened 2019-04-02
Preparations for usage with Accio (https://github.com/JamitLabs/Accio)

siblum:
We at JamitLabs developed a dependency management system relying on the Swift Package Manager and as we included your framework in one of our projects we need to have the Package.swift file, so I added it. For further reading please check https://github.com/JamitLabs/Accio.


robertjpayne:
@siblum this will be good to go with the Swift 5.0 release thanks for your input though!

PR #559, opened 2018-11-16
Add back Package.swift for SwiftPM compatibility

LinusU:
reverts 477fffae4c37386285f78511ff4120f1f5324ff8

SnapKit was actually was compatible with SwiftPM since #501 was merged, this adds the file back

PR #588, opened 2019-04-22
Add Compatibility Version

hlineholm-flir:
This is a basic functionality of a framework and it is required for it to function in a dependency graph. If the compatibility doesn't exist or is set to 0 you'll get:

`
dyld: Library not loaded: @rpath/SnapKit.framework/SnapKit
  Referenced from: /private/var/containers/Bundle/Application/2EAB6DCA-EF3C-461B-97FD-1666A542B2E7/XXX.app/Frameworks/XXX.framework/XXX
  Reason: Incompatible library version: XXX requires version 1.0.0 or later, but SnapKit provides version 0.0.0
`

As you can see, SnapKit tells **dylid** that it is compatible with version 0.0.0 which is not true.


robertjpayne:
@hlineholm @LinusU this needs to be rebased against develop not master otherwise it's hard to decipher the changes at the moment

LinusU:
#559 has already been merged already and no longer needs to be in here :)

Source/ConstraintConstantTarget.swift
```diff
@@ -91,53 +95,73 @@ extension ConstraintConstantTarget {
...
                     return -value.right
-                case .bottom, .bottomMargin:
+                case .bottom, .bottomMargin, .lastBaseline:
                     return -value.bottom
                 case .leading, .leadingMargin:
                     return (ConstraintConfig.interfaceLayoutDirection == .leftToRight) ? value.left : value.right
                 case .trailing, .trailingMargin:
                     return (ConstraintConfig.interfaceLayoutDirection == .leftToRight) ? -value.right : -value.left
+                case .centerX, .centerXWithinMargins:
+                    return (value.left - value.right) / 2
```
colinta: Seems like this should be `(value.left + value.right) / 2` – the average, not the difference...?

colinta: Ohh no I see, the bottom and right values are inverted sign, so this *is* doing addition.  Got it.

robertjpayne:
Believe this is obsolete now

PR #594, opened 2019-05-03
Adds support for NSDirectionalEdgeInsets as an inset constant

jctrouble:
This change adds support for `NSDirectionalEdgeInsets` as a constant target.

Introduced in iOS 11 and tvOS 11, this type behaves similarly to `UIEdgeInsets` with the exception that it accounts for right-to-left rendering.

That is, instead of the `top, left, bottom, right` fields in `UIEdgeInsets`, `NSDirectionalEdgeInsets` has `top, leading, bottom, trailing`.


jctrouble:
@robertjpayne any interest in taking a look at this?

robertjpayne:
👍 looks good to me!

PR #637, opened 2019-10-12
Add Convenience ConstraintItems for dimensional edges

svendr:
This PR adds the ConstraintItems `horizontalEdges`, `verticalEdges`, `directionalHorizontalEdges` and `directionalVerticalEdges` as convenience items to express left+right, top+bottom or leading+trailing.

I've been using these shortcuts for some time now in several projects to conveniently express something like 'Pin this view to the left and right edge of the superview':

```
subview.snp.makeConstraints { make in
    make.horizontalEdges.equalToSuperview()
}
```

This is very useful in cases where you don't want to pin all edges of a view to those of another view using `.edges.equalTo(...)`, but still don't want to type a lot of code to pin specific dimensional edges.

I've also added comprehensive UnitTesting for these new Items.

### DIFF
--- a/Source/ConstraintAttributes.swift
+++ b/Source/ConstraintAttributes.swift
@@ -99,7 +99,11 @@ internal struct ConstraintAttributes : OptionSet, ExpressibleByIntegerLiteral {
     // aggregates
     
     internal static var edges: ConstraintAttributes { return 15 }
+    internal static var horizontalEdges: ConstraintAttributes { return 5 }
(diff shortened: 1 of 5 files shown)


robertjpayne:
Looks great! 👍

pyrou:
Hi @robertjpayne I just wondering, this PR was merged 1 year ago now. 
If I'm correct it still not available in the latest snapkit release ? Did I missed something ?

bixcorp:
Same for me. Looks great but not available in last cart. 😯

PR #645, opened 2019-12-09
Add back support for OS X 10.11

branchseer:
SnapKit works well in OS X 10.11. This PR prevents Xcode from complaining when SnapKit is added to a project whose macOS Deployment Target is 10.11.


Overcot (approved):

Mojo12525 (commented):
I am reallyy starting to think I am a little bit of a troll but this time around it seems to be just me struggling

Issue #630, opened 2019-09-17
constraint(equalToSystemSpacingBelow:multiplier:) not supported


kevinmbeaulieu:
### New Issue Checklis

* [x] I have looked at the [Documentation](http://snapkit.io/docs)
* [x] I have read the [F.A.Q.](http://snapkit.io/faq)
* [x] I have filled out this issue template.

### Issue Info

 Info                    | Value                               |
-------------------------|-------------------------------------|
 Platform                | ios/tvos
 Platform Version        | All
 SnapKit Version         | 5.0.1
 Integration Method      | All
 

### Issue Description

When laying out a series of labels vertically, [constraint(equalToSystemSpacingBelow:multiplier:)](https://developer.apple.com/documentation/uikit/nslayoutyaxisanchor/2866022-constraint) is useful, but it is not currently supported in SnapKit as far as I can tell. While it is possible to create this constraint using vanilla UIKit and the rest of our constraints in SnapKit, it would be ideal if we could do something like
```
label2.snp.makeConstraints { make in 
    make.firstBaseline.equalToSystemSpacingBelow(label1.snp.lastBaseline).multipliedBy(1)
}
```
which would be analogous to
```
label2.firstBaselineAnchor.constraint(equalToSystemSpacingBelow: label1.lastBaselineAnchor, multiplier: 1).isActive = true
```

If this is a feature people would be open to adding, I can look into what it would take to support it.

robertjpayne:
Ah I closed your PR but I see this actually has a system API. Unfortunately it doesn't look like there is any way to get that "standard spacing" without the visual format parsing? Due to the inner workings of SnapKit and remake/update constraints I worry about that actually working that well?

PR #649, opened 2019-12-21
Add equalToSystemSpacingBelow relatable

kevinmbeaulieu:
Enable creation of constraints that use standard system spacing.

Implementation of this enhancement request: https://github.com/SnapKit/SnapKit/issues/630

### DIFF
--- a/Sources/ConstraintMakerRelatable.swift
+++ b/Sources/ConstraintMakerRelatable.swift
@@ -73,6 +73,25 @@ public class ConstraintMakerRelatable {
         editable.description.constant = constant
         return editable
     }
+
+    @available(iOSApplicationExtension 11.0, *)
+    @discardableResult
+    public func equalToSystemSpacingBelow(_ other: ConstraintItem, _ file: String = #file, _ line: UInt = #line) -> ConstraintMakerEditable {
+        let editable = ConstraintMakerEditable(self.description)
+        editable.description.sourceLocation = (file, line)
+        editable.description.relation = .equal
+        editable.description.related = other
+        editable.description.constant = LayoutConstraint.constraints(
+            withVisualFormat: "V:[topView]-[bottomView]",
+            options: .spacingBaselineToBaseline,
+            metrics: nil,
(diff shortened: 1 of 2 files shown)


kevinmbeaulieu:
Hm can't seem to get macOS to build. Looks like `NSLayoutConstraint.FormatOptions.spacingBaselineToBaseline` is only supported on macOS starting with Mac Catalyst 13+, but I'm unable to find a `@available` annotation that successfully guards the logic. If anyone has ideas lmk and I can try, otherwise will close this out 😞

kevinmbeaulieu:
cc @robertjpayne since it looks like you're by far the heaviest contributor here

kevinmbeaulieu:
Looks like iOS build is failing for unrelated issue, put up a fix for that here https://github.com/SnapKit/SnapKit/pull/650

CognitiveWin:
Just curious, how’s this looking?

kevinmbeaulieu:
@CognitiveWin Should be ready for review, if @robertjpayne can take a look

kevinmbeaulieu:
@robertjpayne Any thoughts on this? I can close out if you're not interested in this addition

philipbel:
@robertjpayne Can you please review or provide input on this PR?
It would be a great addition to SnapKit.
Thanks.

robertjpayne:
These sorts of API extensions are still best left to your own app level extensions to prevent API bloat. I understand they are convenience but every organisation has their own set of best practice here and there's not a well defined API from Apple to copy or make something similar from. 

They're also going to be possible with the opening of the ConstraintItem API in #693

notapplicableio:
I disagree.

I think this is not bloat but within reason when it provides a mapping to a well defined method provided by Apple in https://developer.apple.com/documentation/uikit/nslayoutyaxisanchor/2866022-constraint (the same area of functionality under the XY anchors constraints). 

The consequence of this decision is the client code-app suffers bloat to add an extension for an NSLayout to an otherwise complete SDK.

PR #650, opened 2019-12-27
Update Travis to build for iPhone 8

kevinmbeaulieu:
After updating to Xcode 11, the iPhone 6S simulator is no longer
available, causing all iOS builds to fail.


kevinmbeaulieu:
@robertjpayne are you still maintaining SnapKit or should folks who still want to use it start maintaining their own fork?

kevinmbeaulieu:
Looks like https://github.com/SnapKit/SnapKit/pull/645 got the build working again, so this isn't needed anymore

PR #673, opened 2020-06-12
Update README to include SPM installation notes

kwakueshun:
This PR adds installation notes for adding SnapKit to a project using Swift Package Manager. I recently moved my project to use the SPM version of Snapkit to my project smoothly and thought it might be worth adding this for future visitors.


robertjpayne:
Thanks 👍

PR #674, opened 2020-07-03
Add constraints to an array of views

snail-z:
```swift
let views = [view1, view2, view3]
views.snp.distributeViewsAlong(axis: .horizontal, fixedSpacing: 10, leadSpacing: 10, tailSpacing: 10)
views.snp.makeConstraints { (make, element) in
    make.top.equalTo(100)
    make.height.equalTo(element.snp.width)
}
```


tanpengsccd:
Dose this function implement?

robertjpayne:
These sorts of API extensions are still best left to your own app level extensions to prevent API bloat. I understand they are convenience but every organisation has their own set of best practice here and there's not a well defined API from Apple to copy or make something similar from.

Issue #668, opened 2020-05-18
How to use aspect ratio constraint?


rafagan:
### New Issue Checklist

🚫 If this template is not filled out your issue **will** be closed with no comment. 🚫

* [X] I have looked at the [Documentation](http://snapkit.io/docs)
* [X] I have read the [F.A.Q.](http://snapkit.io/faq)
* [X] I have filled out this issue template.

### Issue Info

 Info                    | Value                               |
-------------------------|-------------------------------------|
 Platform                | ios
 Platform Version        | 13.0
 SnapKit Version         | latest
 Integration Method      | cocoapods
 

### Issue Description

I trying to create an aspect ratio constraint 1:1. How can I achieve that?

rafagan:
In swift I've done in that way:
`
extension UIView {
    func aspectRatio(_ ratio: CGFloat) -> NSLayoutConstraint {
        return NSLayoutConstraint(item: self, attribute: .height, relatedBy: .equal, toItem: self, attribute: .width, multiplier: ratio, constant: 0)
    }
}
`

In snapkit:

`
button.snp.makeConstraints { (make) in
                make.width.equalTo(button.snp.height).multipliedBy(1.0 / 1.0)
            }
`

robertjpayne:
Will be possible in next release via your own extensions due to #693

Issue #417, opened 2017-04-17
Feature request: add aspect ratio "helper" method


fcy:
I've seen some old issues with questions about aspect ratio and if I understood correctly the reason there no SnapKit method for it is because there is no equivalent API on AutoLayout. My question is are you against adding a method like the following (naming TBD)?

```
profileImageView.snp.makeConstraint { make in 
    make.aspectRatio(1, by: 1) // square. or aspectRatio(16, by: 9)
}
```

that would internally be the same as:

```
profileImageView.snp.makeConstraint { make in 
    make.width.equalTo(profileImageView.snp.height).multipliedBy(1 / 1)
}
```

The benefit is that it makes the intent of the constrain very clear.

If this is something you'd add to the project I can create a PR.

robertjpayne:
@fcy It's an interesting idea, I'm sort of against bloating SnapKit's API's with too much convenience, while useful I don't think there are that many instances where one needs to do aspect ratio pinning to warrant a top level API.

Instead it's easy to add extensions to protocols/classes that DSL works from and keep it application specific.

It may be a case where I need to add a `Contrib` folder with additional API's and helpers that are not part of the core package.

robertjpayne:
@fcy my take on it is that I want SnapKit to closely mirror Auto Layout. Because if one comes into a new project that uses SnapKit and the original author is gone if there are tons of helpers it's even more confusing than just starting with the basics.

Most of SnapKit's API's read well with their intention `make left equal to right offset 100` etc… and this is a core goal because layout code is verbose and becomes lengthy and hard to understand visually when reading.

fcy:
That makes sense and I had a feeling this was the way the project wanted to go, that is why I started by asking about it. I ended up adding an extension to `ConstraintMaker`:

```
extension ConstraintMaker {
    public func aspectRatio(_ x: Int, by y: Int, self instance: ConstraintView) {
        self.width.equalTo(instance.snp.height).multipliedBy(x / y)
    }
}
```

```
        profileImageView.snp.makeConstraints { make in
            make.aspectRatio(1, by: 1, self: profileImageView)
        }
```

The only thing that bothers me is that I have to pass a reference to the view itself, I couldn't find a way to reference the view from the maker.

Any tips on how this could be done better?

dev4dev:
@fcy maybe something like this?

``` swift
extension ConstraintMaker {
	public func aspectRatio(_ x: Int, by y: Int) {
		if let view = item as? ConstraintView {
[...]

Issue #107, opened 2015-06-04
Aspect Ratio Constraint to Docu


confile:
Hi, 

could you please add the aspect ration constraint to the documentation. I see this in one of the closed issues. I think this would help a lot.

confile:
Also, may be an example of equal width constraint.

robertjpayne:
Will be possible in next release via your own extensions due to #693

Issue #92, opened 2015-05-15
Aspect Ratio Constraint?


sleroux:
Hey there,

I can't find a way to set a view's aspectRatio constraint property using Snapkit. Is this supported?

robertjpayne:
Hey @sleroux is this for an OS X application? or iOS? I'm not aware of any aspectRatio API's for auto layout?

sleroux:
Sorry my bad. IB has an aspectRatio constraint as part of it's UI but there isn't one that exists through the code API.

robertjpayne:
Will be possible in next release via your own extensions due to #693

PR #693, opened 2020-12-07
Make ConstraintMaker.item in public access level

### LINKED ISSUE #666: would you change ConstraintMaker.item to public?
### New Issue Checklist

* [x] I have looked at the [Documentation](http://snapkit.io/docs)
* [x] I have read the [F.A.Q.](http://snapkit.io/faq)
* [x] I have filled out this issue template.

### Issue Info

 Info                    | Value                               |
-------------------------|-------------------------------------|
 Platform                | iOS 
 Platform Version        | 10
 SnapKit Version         | 5.0.1
 Integration Method      | cocoapods
 

### Issue Description
I am trying to extend ConstraintMaker to restrict the View's height and width with it's original ratio, like:  
```
extension ConstraintMaker {
	@inline(__always)
	public func combineWidthToHeight(ratio: CGFloat? = nil) {
		guard let view = item as? UIView else { return }
		let ratio = ratio ?? (view.frame.size.width / view.frame.size.height)
		width.equalTo(view.snp.height).multipliedBy(ratio)
	}
}
```
and I trying to make more extension about ConstraintMaker for my convenience.

But the question is, ConstraintMaker.item is private, so I have change the source code so that I can use it, would you please change ConstraintMaker.item to public? It looks like no harm to use public. I am really don't want to change the source code, and I really love SnapKit, Thanks.

yunhao:
The item property of ConstraintMaker is now in public access level. Developers can easily extend ConstraintMaker to add features.

There are many issues discussing about how to make aspect ratio constraint. However, because of the internal access level, developer can't create extension to extend SnapKit.

Now, we can do this:
```swift
extension ConstraintMaker {
    public func aspectRatio(_ ratio: CGSize) {
        let view = item as! ConstraintView
        self.width.equalTo(view.snp.height).multipliedBy(ratio.width / ratio.height)
    }
}
```

With this change, developer can create any helper method they like. They don't have to create issues to request this kind of features.

Related issue: #668, #417, #107, #92
Close #666

### DIFF
--- a/Sources/ConstraintMaker.swift
+++ b/Sources/ConstraintMaker.swift
@@ -163,7 +163,7 @@ public class ConstraintMaker {
         return self.makeExtendableWithAttributes(.centerWithinMargins)
     }
     
-    private let item: LayoutConstraintItem
+    public let item: LayoutConstraintItem
     private var descriptions = [ConstraintDescription]()
     
     internal init(item: LayoutConstraintItem) {


robertjpayne:
I'll accept this. I never wanted to open this API because it will cause more issues in the tracker of people doing weird things but I agree it'd be super nice to be able to create your own powerful extensions.

Sorry it took forever for me to add this in!

svycka:
> I'll accept this. I never wanted to open this API because it will cause more issues in the tracker of people doing weird things but I agree it'd be super nice to be able to create your own powerful extensions.
> 
> Sorry it took forever for me to add this in!

When this can be released?

Issue #442, opened 2017-09-01
Unable to use SnapKit inside Swift Playground


Krisiacik:
* [x ] I have looked at the [Documentation](http://snapkit.io/docs)
* [x ] I have read the [F.A.Q.](http://snapkit.io/faq)
* [x ] I have filled out this issue template.

### Issue Info

 Info                    | Value                               |
-------------------------|-------------------------------------|
 Platform                | ios
 Platform Version        | 10
 SnapKit Version         | 3.2.0
 Integration Method   | cocoapods
 

### Issue Description

1) I created a simple single-view project called QuickViewBuilder (just a name ..doesn't really matter, but it suggests what I want to acchive - it is to build views, view controllers faster..using swift playgrounds..)
2) I promoted the project to workspace using Cocoapods, having SnapKit as a pod.
3) I created a Playground right inside the QuickViewBuilder project.
4) I build the project .(not ran, just built so that the SnapKit dependency is prebuilt)
5) I created a viewController inside the playground using the playground template in Xcode9. But really all there is to it is this line 
PlaygroundPage.current.liveView = MyViewController()
so you can actually visualize your playground code on the right pane as if it was an app.
6) I imported SnapKit using "import SnapKit" into playground.
7) For now, no SnapKit code used in the controller, only the import declaration and the playground runs properly. 
7) But - as soon as I use a single line of code  trying to use snakpit to lay out subviews, i end up with this.
8) Screnshots attached to prove there is *no other issue* in this whole setup.
9) I can build and render the viewcontroller easily as long as I do not use SnapKit code.
10) If I do, I end up with this error message:

Playground execution failed:

error: Couldn't lookup symbols:
  __T07SnapKit17ConstraintViewDSLVN
  __T07SnapKit23ConstraintMakerEditableCMa
  __T07SnapKit17ConstraintViewDSLV15makeConstraintsyyAA0C5MakerCcF
  __T0So6UIViewC7SnapKitE3snpAC17ConstraintViewDSLVfg

Can you tell me why and what is the fix please?

![nosnapkitworks](https://user-images.githubusercontent.com/9870899/29978185-04ce3696-8f38-11e7-8297-f01ef054dd37.png)
![snapkitnotworking](https://user-images.githubusercontent.com/9870899/29978188-05f00e64-8f38-11e7-8e8e-5e2a99c748b0.png)

robertjpayne:
[...]

PR #699, opened 2020-12-26
fix SnapKitPlayground can not use SnapKit after build SnapKit.framewo…

leavesster:
fix (#442)
I have the same problem as described in #442

I open `SnapKit.xcworkspace` not `SnapKit.xcodeproj`, and build SnapKit. and try as the README said, but it not work.
>To try SnapKit in playground, open SnapKit.xcworkspace and build SnapKit.framework for any simulator first

I read this blog [Using 3rd party frameworks in Xcode 10 Playgrounds](https://medium.com/@pardel/using-3rd-party-frameworks-in-xcode-10-playgrounds-97aa0c3f56e0) , and find that the SnapKitPlayground is not in same project. so I rebuilt a new identical file.

I just change the playground's project belong to, it seems work fine to me.

here is the playground create config, I just add the playground to `SnapKit` project.
create

project

And after that the `SnapKitPlayground.playground` is under `SnapKit` project. And I found After this, the `SnapKit.xcworkspace` is no long needed. Just open `SnapKit.xcodeproj` and build framework is OK to play this playground

### DIFF
--- a/SnapKit.xcodeproj/project.pbxproj
+++ b/SnapKit.xcodeproj/project.pbxproj
@@ -50,6 +50,7 @@
 
 /* Begin PBXFileReference section */
 		2DBA080D1F1FAD66001CFED4 /* Typealiases.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Typealiases.swift; sourceTree = "<group>"; };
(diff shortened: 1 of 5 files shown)


teameh:
Still same thing for me on your branch

PR #716, opened 2021-05-07
Resolve Xcode 12.5 warning regarding `class` protocols

pouriaalmassi:
Resolve Xcode 12.5 warning regarding using the class keyword being deprecated in favor of AnyObject for protocol inheritance.


pouriaalmassi:
Duplicates #714

PR #720, opened 2021-05-26
Update docs to reflect current api.

Journey36:
I think this docs was written by mixing `Objective-C` style and `Swift` style.
So, I changed part of docs to `Swift` style and current API for overall unity.
In addition, I removed part of docs that deprecated. Because it can confuse the person who reading the docs.


robertjpayne:
Thank you!

PR #724, opened 2021-08-26
Implement UIView addSubview + makeConstraints extension

ScARje:
Extension for `UIView` and `NSView` that adds new `addSubview` function which takes `(ConstraintMaker) -> Void` closure as an argument, calls default addSubview function and applies closure to the added view.

As a result, code that used to look like this:

```swift
addSubview(myView)
myView.makeConstraints {
...
}
```

Will look like this:
```swift
addSubview(myView) {
...
}
```

Based on my experience of using SnapKit, adding subview then applying constraints to it is one of the most common patterns in screen-building code. So potentially this little extension might reduce the number of code lines and slightly improve code readability.

### DIFF
--- a/SnapKit.podspec
+++ b/SnapKit.podspec
@@ -12,7 +12,7 @@ Pod::Spec.new do |s|
   s.osx.deployment_target = '10.11'
   s.tvos.deployment_target = '10.0'
 
-  s.source_files = 'Source/*.swift'
+  s.source_files = 'Sources/*.swift'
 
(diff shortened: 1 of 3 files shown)


robertjpayne:
I appreciate your input here -- however as part of the direction of SnapKit is to minimise any sort of extension of System API's as much as possible (hence nearly everything is namespaced under `.snp.`) I'm going to close this for now.

These extensions are easily available via your own extensions in your apps though and I recommend to keep doing that.

PR #737, opened 2021-11-09
fix the logic of `prepareConstraints` to pass through compactMap

toya108:
I thought it more efficient to use the `compactMap` to exclude nil cases first, rather than checking all cases in the for loop.


robertjpayne:
@toya108 are you actually sure about this? In my experience using closure based API's (map, compactMap etc…) are slower than for loops in Swift.

toya108:
Thank you.
I'm sorry. I wasn't sure, I just had a theory.
I have investigated the speed of the for loop and the closure-based API.
There are differences in the type of API and whether it is optimized or not, but it seems to be faster with for loop.
Therefore, since there is no proof that this PR will make the process more efficient, I close this PR.

PR #750, opened 2022-04-14
Add missing 'Superview' word in ConstraintMakerRelatable extension method name

mikchmie:
One of the methods added in https://github.com/SnapKit/SnapKit/pull/663 is missing the 'Superview' word in its name.


jzdeng:
你好，我已经收到您发来的邮件，祝您身体健康，万事如意^_^     -邓江洲

robertjpayne:
@mikchmie good catch sorry it took me forever to see this

PR #763, opened 2022-09-20
fix for Xcode 14 : iOS 12 crash dyld: Library not loaded: /usr/lib/swift/libswiftCoreGraphics.dylib Reason: image not found

Dwarven:
fix for Xcode 14 : iOS 12 crash 
dyld: Library not loaded: /usr/lib/swift/libswiftCoreGraphics.dylib
Reason: image not found


atsumori:
atsumori:hotfix/xcode14

PR #793, opened 2023-12-26
Add Privacy Manifest

wlxo0401:
I tried adding PrivacyInfo. I'm not sure if it's done right because I've never done PR before.

Below is a document listing API that Apple needs to add description. I checked and it didn't seem to be the case for Snapkit. 
[Describing use of required reason API](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api)
I would appreciate it if someone could double-check my findings to confirm their accuracy.

And I revised 'Package.swift' and 'SnapKit.podspec' by referring to the libraries that have already added PrivacyInfo.

But I'm not sure if this is the right thing to do. I want everyone to check it out together.

If you see anything that needs to be corrected, please share it with us.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.0
+// swift-tools-version:5.3
 //
 //  SnapKit
 //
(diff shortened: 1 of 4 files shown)


robertjpayne:
I missed this but have just published 5.7.0 that should work for direct and/or SwiftPM integrations

PR #798, opened 2024-01-12
Include privacy info for Cocoapods

dcacenabes:
Recently SnapKit introduced support for Privacy manifest but only for SPM. This PR introduces the support for Cocoapods based projects by adding using resource_bundles in the podspec.
While using `resources` can lead to collisions for resources with the same name, using `resource_bundles` is safe because it is the bundle that gets copied to the main app target, not the contents of the bundle themselves.

More details on the solution using resource_bundles can be found in this great comment in the Google promises repository, which is working on the same fix:

https://github.com/google/promises/pull/226#discussion_r1447871477

### DIFF
--- a/SnapKit.podspec
+++ b/SnapKit.podspec
@@ -20,5 +20,9 @@ Pod::Spec.new do |s|
       'LIBRARY_SEARCH_PATHS' => '$(SDKROOT)/usr/lib/swift',
   }
 
+  s.resource_bundles = {
+    'SnapKit_Privacy' => ['Sources/PrivacyInfo.xcprivacy'],
+  }
+
   s.swift_versions = ['5.0']
 end


wlxo0401:
@dcacenabes 

I'm wondering if it shouldn't be 
'SnapKit' => ['Sources/PrivacyInfo.xcprivacy'] instead of 'SnapKit_Privacy' => ['Sources/PrivacyInfo.xcprivacy']. 

Looking at other libraries, it seems that resources are set using the library name... I might be wrong, but I think it's worth checking.

robertjpayne:
I'm holding out on this until there is a bit more official response from the Cocoapods teams, though I'm not sure there will be because the project seems to be mostly in stable and maintenance mode (which is totally fair!).

I'm not sure that this method of including it covers all use cases of Cocoapods but I'm willing to merge it if nothing further emerges from their end by the first weekish in Feb.

wlxo0401:
> I'm holding out on this until there is a bit more official response from the Cocoapods teams, though I'm not sure there will be because the project seems to be mostly in stable and maintenance mode (which is totally fair!).
> 
> I'm not sure that this method of including it covers all use cases of Cocoapods but I'm willing to merge it if nothing further emerges from their end by the first weekish in Feb.

Apple enforces Privacy Manifest, but doesn't seem to share the curiosity clearly in applying it 😭

It would be nice to have a lot of information..😢

dcacenabes:
> @dcacenabes
> 
> I'm wondering if it shouldn't be 'SnapKit' => ['Sources/PrivacyInfo.xcprivacy'] instead of 'SnapKit_Privacy' => ['Sources/PrivacyInfo.xcprivacy'].
> 
> Looking at other libraries, it seems that resources are set using the library name... I might be wrong, but I think it's worth checking.

Hey @wlxo0401 .
You can define whatever you want as a bundle name. I was following this approach of adding the _Privacy to the bundle for 2 reasons mainly:
- After bumping to Xcode 15 we started having issues with bundles that had the same name as the module when loading assets in test targets. Changing the name to something different than the module name fixed it.
- Google is also taking this approach to adding _Privacy to the bundle name: https://github.com/google/promises/pull/226/files#diff-dd80f678cfd0518f567236e913b53c5ff56dffff5eb81891d04f6c440d1d7d21R27 

Unless the maintainer of this library has a strong opinion against the current naming I would like to keep it as is.

dcacenabes:
> I'm holding out on this until there is a bit more official response from the Cocoapods teams, though I'm not sure there will be because the project seems to be mostly in stable and maintenance mode (which is totally fair!).
> 
> I'm not sure that this method of including it covers all use cases of Cocoapods but I'm willing to merge it if nothing further emerges from their end by the first weekish in Feb.

Thanks a lot for your reply @robertjpayne ! We will keep an eye here to any comms from Cocoapods or Apple regarding this matter.

dcacenabes:
Hi @robertjpayne !
It is almost mid-February already and there are no news from Cocoapods. Would it be possible to merge this already so that we can start adopting the latest version of the library? Thanks in advance!

robertjpayne:
@dcacenabes thanks :) Will publish to trunk tonight hopefully

robertjpayne:
@wlxo0401 @dcacenabes also I agree `_Privacy` on the bundle suffix. Too many things may make for weird collisions leaving it the same as the library itself. Also it's basically hidden and irrelevant to developer or end users.

dcacenabes:
Thanks @robertjpayne !
Looking forward to it getting released to start shipping this in our newest build :D !

dimpy-iroid:
Hello @robertjpayne, does latest version of SnapKit includes "PrivacyInfo" file? We are getting email from the Apple:

ITMS-91061: Missing privacy manifest - Your app includes “Frameworks/SnapKit.framework/ SnapKit”, which includes SnapKit, an SDK that was identified in the documentation as a commonly used third-party SDK. Starting February 12, 2025, if a new app includes a commonly used third- party SDK, or an app update adds a new commonly used third-party SDK, the SDK must include a privacy manifest file or you won’t be able to submit the app for review in App Store Connect. Please contact the provider of the SDK that includes this file to get an updated SDK version with a privacy manifest. For more details about this policy, including a list of SDKs that are required to include signatures and manifests, visit: https://developer.apple.com/support/third-party-SDK-requirements.

We have to upload the build on the TestFlight and then submit it for the Apple Review.

cemcemi (commented):
cd9d05bd7e72ac5fd5d0b8fdfdc885d043e109f6

Mojo12525 (commented):
Love this

Mojo12525 (commented):
- [ ] adj3707 loves this **

jesn459 (commented):
Gracias

jesn459 (commented):
Gracias

jesn459 (commented):
Gracias

jesn459 (commented):
Gracias

jesn459 (commented):
Gracias

jesn459 (commented):
Gracias

PR #812, opened 2024-07-07
Updated Example-iOS deployment target

akrzywicki:
This fix raises the minimum target of the test project and allows the Example-iOS project to be built correctly.


Naveen-C-Ramachandrappa (approved):
Thanks for updating this.

PR #823, opened 2024-12-16
Update #file to #fileID to avoid leak full path information

Kyle-Ye:
## Motivation

`#file` will have the same effect of `#filePath` in Swift 5 language mode and `#fileID` in Swift 6 language mode.

> Because #fileID doesn’t embed the full path to the source file, unlike #filePath, it gives you better privacy and reduces the size of the compiled binary. Avoid using #filePath outside of tests, build scripts, or other code that doesn’t become part of the shipping program.

In most cases `#fileID` should be preferred as it has better privacy and reduces the binary size.

example:

- `filePath`: `/Users/ci/build/projectX/AKit/UI/Manager.swift`
- `fileID`: `AKit/Manager.swift`

## Ref
- https://developer.apple.com/documentation/swift/filepath()
- https://developer.apple.com/documentation/swift/fileid()
- https://github.com/swiftlang/swift-evolution/blob/main/proposals/0285-ease-pound-file-transition.md
- https://forums.swift.org/t/file-vs-fileid-in-swift-6/74614

### DIFF
--- a/Sources/ConstraintMakerRelatable+Extensions.swift
+++ b/Sources/ConstraintMakerRelatable+Extensions.swift
@@ -31,7 +31,7 @@
 extension ConstraintMakerRelatable {
   
     @discardableResult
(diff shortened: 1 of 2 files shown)


robertjpayne:
This looks great! Thank you 👏🏻

